Issue #34, opened 2026-02-17
Renpho ES-CS20M scale does not work


Tosiman-Global:
## What happened?
When I ran the configuration wizard the scale didn't auto discover. So I put in the mac address of the scale. After I everything was configured I ran `npm start`. The program managed to find the scale but couldn't connect to it. Here are the logs. 

## What I expected?

- auto discovered by wizard
- The data from the scale gets pulled to MQTT

```
> ble-scale-sync@1.3.0 start
> tsx src/index.ts

2026-02-17 13:57:46.004 [Config] Loading config from config.yaml

2026-02-17 13:57:46.015 [Sync] BLE Scale Sync
2026-02-17 13:57:46.015 [Sync] Scanning for scale ED:67:39:DC:60:60...
2026-02-17 13:57:46.015 [Sync] Adapters: Senssun Fat Scale, QN Scale, Renpho ES-WBE28, Renpho ES-26BB, Xiaomi Mi Scale 2, Yunmai, Beurer / Sanitas, Sanitas SBF72/73, Soehnle Shape/Style, Medisana BS44x, Trisa, ES-CS20M, Exingtech Y1, Excelvan CF369, Hesley, Inlife, Digoo, 1byone (Eufy), 1byone Scale (new), Active Era BS-06, MGB (Swan/Icomon/YG), Hoffen BS-8107, Standard GATT (BCS/WSS)

2026-02-17 13:57:46.015 [Sync] Running exporter healthchecks...
2026-02-17 13:57:46.289 [Sync]   mqtt: OK
2026-02-17 13:57:46.348 [BLE] Scanning for device...
2026-02-17 13:57:56.373 [BLE] Still scanning...
2026-02-17 13:58:06.403 [BLE] Still scanning...
2026-02-17 13:58:16.455 [BLE] Still scanning...
2026-02-17 13:58:26.505 [BLE] Still scanning...
2026-02-17 13:58:36.542 [BLE] Still scanning...
2026-02-17 13:58:46.582 [BLE] Still scanning...
2026-02-17 13:58:56.362 [BLE] Connect error: could not connect to device: result is null. Retrying (1/5) in 1000ms...
2026-02-17 13:58:57.379 [BLE] Connect error: could not connect to device: result is null. Retrying (2/5) in 1500ms...
2026-02-17 13:58:58.881 [BLE] Connect error: could not connect to device: result is null. Retrying (3/5) in 2000ms...
```
## Steps I have tried to do

[...]

PR #45, opened 2026-02-22
ESP32 (and more) BLE Proxy

APIUM:
## Summary

Adds support for using a BLE proxy device instead of relying on the server directly so you can host on any device.

The proxy is built with Micropython, so there is support for multiple devices. I have been targeting an ESP32-S3 with a 480x480 display on it, so have also added an optional LVGL display component which shows that it's connected, and what the status of the reading is when it detects one. 

I've put a bit of work into the documentation and structure so it shouldn't be to difficult to add support for more devices, though new screens will require work with LVGL.

I've been testing this with a broadcast scale, so some assistance to test with GATT scales would be appreciated. I can assist with adding a board profile or two for community members if you have devices laying around. Make sure it can have both WiFi and BLE radios on at the same tie, they don't have to simultaneously transmit, but the stability is average when we keep having to turn the radios off to transmit as we can miss scale readings. 

## Test Plan

- [x] Tests pass (`npm test`)

Tested with my Renpho ES-CS20M and this device https://www.aliexpress.com/item/1005008214679682.html

### DIFF
--- a/src/ble/handler-mqtt-proxy.ts
+++ b/src/ble/handler-mqtt-proxy.ts
@@ -507,6 +507,7 @@ export class ReadingWatcher {
   private gattInProgress = false;
   private gattStartedAt = 0;
   private _client: MqttClient | null = null;
+  // eslint-disable-next-line @typescript-eslint/no-explicit-any
   private _lifecycleHandlers: Array<{ event: string; handler: (...args: any[]) => void }> = [];
   private _messageHandler: ((topic: string, payload: Buffer) => void) | null = null;
   private _subscribedTopics: string[] = [];
--- a/src/scales/qn-scale.ts
+++ b/src/scales/qn-scale.ts
@@ -7,7 +7,6 @@ import type {
   BodyComposition,
 } from '../interfaces/scale-adapter.js';
 import { uuid16 } from './body-comp-helpers.js';
-import type { ScaleBodyComp } from './body-comp-helpers.js';
 
 /**
  * Ported from openScale's QNHandler.kt
@@ -50,13 +49,6 @@ const CHR_WRITE_T1 = uuid16(0xffe3);
 // Service UUIDs for matching
 const SVC_T1 = 'ffe0';
 const SVC_T2 = 'fff0';
-
-// QN broadcast protocol marker bytes
-const QN_MARKER_0 = 0xaa;
-const QN_MARKER_1 = 0xbb;
-
-/** Minimum length of a valid QN broadcast advertisement payload. */
-const QN_BROADCAST_MIN_LEN = 26;
 
 export class QnScaleAdapter implements ScaleAdapter {
   readonly name = 'QN Scale';


KristianP26 (commented):
## Code Review — PR #45: ESP32 BLE MQTT Proxy

Great work on this, @APIUM. This is a massive feature addition and the architecture is solid — MQTT as a transport layer for BLE is the right approach for headless servers. Below is a detailed review covering Node.js, firmware, and overall structure.

---

### PR Size & Structure

**This PR is too large to review effectively** (61 files, +7,578 lines). I'd strongly recommend splitting it into separate PRs:

1. **Core handler** — `handler-mqtt-proxy.ts`, interface changes (`scale-adapter.ts`), `shared.ts` integration, config schema, tests
2. **Firmware** — `firmware/`, `drivers/`, `PORTING.md`
3. **Docker** — Dockerfile (Alpine migration, multi-stage build), compose files
4. **Docs** — `docs/guide/esp32-proxy.md`, VitePress config

This makes each piece reviewable and mergeable independently.

---

### Node.js — `handler-mqtt-proxy.ts`

**Overall:** Well-structured. The `AsyncQueue`, GATT-over-MQTT abstraction (`MqttBleChar`/`MqttBleDevice`), and the reuse of `waitForRawReading()` from `shared.ts` are all clean.

**Issues:**

1. **Module-level mutable state** — `_persistentClient`, `discoveredScaleMacs`, `_displayUsers` are module globals. This prevents running multiple proxy devices simultaneously and makes testing fragile (requires `_reset*()` helpers). Consider encapsulating state in a class (e.g., `MqttProxyManager`) or passing it through function params.

2. **`ReadingWatcher` listener leak** — In `start()`, event listeners (`reconnect`, `offline`, `error`, `connect`) are attached to the persistent MQTT client but never removed. If the watcher is recreated (e.g., config reload), these accumulate on the shared client.

3. **`gattInProgress` flag has no timeout** — If `handleGattReading()` hangs (MQTT broker goes down mid-GATT, ESP32 crashes), `gattInProgress` stays `true` forever and all future GATT readings are silently skipped. Add a timeout or auto-reset mechanism.

4. **MQTT topics never unsubscribed** — `subscribeAsync()` is called in many places (`mqttScan`, `mqttGattConnect`, `MqttBleChar.subscribe`, etc.) but `unsubscribeAsync()` is never called. For ephemeral clients this is fine (disconnect cleans up), but for the persistent client, subscriptions accumulate across scan cycles.

5. **`MqttBleChar.read()` doesn't clean up subscription on timeout** — When `withTimeout` rejects, the response topic remains subscribed and the handler stays attached (only removed on success). The `catch` block removes the listener but doesn't unsubscribe from the topic.

6. **`dummyProfile` in `ReadingWatcher`** — Falls back to a hardcoded `{ height: 170, age: 30, gender: 'male', isAthlete: false }` for GATT readings when no profile is provided. This will produce incorrect body composition. Either make `profile` required or log a clear warning.

**Nits:**
- `scanAndReadRaw` creates an ephemeral client but `ReadingWatcher` uses the persistent one — the dual-path client management works but is easy to get wrong. Consider documenting which functions use which client and why.
- The dedup key `${address}:${weight.toFixed(1)}` means a 0.05kg difference (72.45 vs 72.50) could pass dedup. Intentional?

---

### Node.js — `renpho-broadcast.ts`

**This duplicates existing logic.** The AABB broadcast protocol (company ID `0xFFFF`, bytes 0-1 magic `0xAABB`, byte 15 stability `0x25`, bytes 17-18 LE uint16 weight / 100) is identical to the broadcast parsing we already have in `qn-scale.ts` on the `dev` branch (fixed in #34).

**Recommendation:** Instead of a separate `RenphoBroadcastAdapter`, add `parseBroadcast()` to the existing `QnScaleAdapter`. The QN adapter already matches Renpho devices by name. This avoids two adapters fighting over the same device in the adapter matching loop.

If you want to keep them separate for clarity, at minimum extract the shared AABB parsing into a helper function used by both.

---

### Node.js — `scale-adapter.ts` interface changes

The `parseBroadcast?()` optional method addition is a good API design — clean opt-in for broadcast-capable adapters. The `BleDeviceInfo.manufacturerData` field was already there from earlier work, so this integrates well.

---

### Firmware — `ble_bridge.py`

**Overall:** Solid. Good choices throughout.

1. **Raw IRQ scanning over aioble's scanner** — Correct choice. aioble doesn't expose raw manufacturer data reliably, and the manual AD structure parsing is necessary for broadcast scale detection. The parser handles AD types 0x08/0x09 (name), 0x02/0x03 (16-bit UUIDs), and 0xFF (manufacturer) correctly.

2. **UUID normalization** (`_norm_uuid`) — Correctly mirrors Node.js `normalizeUuid()`. This is critical for characteristic matching across the MQTT bridge. Good.

3. **Dedup in `scan()`** — Keeps strongest RSSI, updates manufacturer data from later advertisements. Correct behavior for broadcast scales where you receive multiple ADVs.

4. **`start_notify` disconnect detection** — The `_disconnect_fired` flag ensures the callback fires exactly once. Good defensive coding.

**Issues:**

5. **`scan()` `MAX_SCAN_ENTRIES` hard cap** — If the cap is hit, later (possibly more relevant) results are silently dropped. Consider logging when the cap is reached.

6. **`connect()` service/characteristic discovery has no timeout** — `device.connect(timeout_ms=15000)` is good, but the subsequent `services()` and `characteristics()` discovery has no timeout. On flaky BLE connections, this could hang.

7. **`write()` auto-detects response mode** — Uses `FLAG_WRITE` to decide `response=True/False`. Some scales require write-without-response even when the characteristic technically supports write-with-response. This may need a way to override per-adapter, but fine for now.

---

### Firmware — `main.py`

**Overall:** Clean architecture. The sync callback -> `_pending` queue -> async loop pattern is exactly right for MicroPython.

1. **`board.DEACTIVATE_BLE_AFTER_SCAN`** — Well-handled. The WiFi recovery loop (wait for `mqtt_as` to reconnect, with `_subs_ready` gating) is necessary for shared-radio ESP32 boards and seems reliable.

2. **Scale MAC beep detection** — 60s debounce via `_last_beep_time` is reasonable.

**Issues:**

3. **Screenshot handler hardcodes 480x480 RGB565** — Should read dimensions from the board abstraction (`board.DISPLAY_WIDTH`, `board.DISPLAY_HEIGHT`) rather than hardcoding. Other boards may have different resolutions.

4. **`handle_connect` busy-wait** — The 30s busy-wait (`for _ in range(60): if not _busy: break; await asyncio.sleep_ms(500)`) is functional but coarse. Consider using an `asyncio.Event` instead.

5. **`_pending.pop(0)`** — O(n) list pop from front. Fine for small queues, but `collections.deque` would be more idiomatic if available in MicroPython.

---

### Docker Changes

1. **`docker-compose.yml` committed directly** — This will confuse users who use direct BLE (which uses `docker-compose.example.yml`). Rename to `docker-compose.mqtt-proxy.yml` or `docker-compose.mqtt-proxy.example.yml` to avoid ambiguity.

2. **Alpine migration** — Switching from `node:20-slim` (Debian) to `node:20-alpine` is a good size improvement, but this is a potentially breaking change for existing users. BlueZ on Alpine can have subtle differences. This should be called out in release notes.

3. **Multi-stage build** — Good improvement. Compiling TypeScript in the build stage and copying `dist/` to runtime is cleaner and produces smaller images.

---

### Summary

| Area | Verdict |
|------|---------|
| Architecture (MQTT proxy concept) | Excellent — right approach |
| `handler-mqtt-proxy.ts` | Good, needs state management cleanup |
| `renpho-broadcast.ts` | Duplicate — consolidate with `qn-scale.ts` |
| `ble_bridge.py` | Solid |
| `main.py` | Clean, minor improvements needed |
| Docker changes | Good but needs naming fix |
| **PR size** | **Too large — split recommended** |

The core functionality is sound and well-tested (1,243 lines of handler tests!). Main asks: consolidate the Renpho adapter with QN, fix the state management issues in the handler, rename docker-compose.yml, and ideally split this into smaller PRs for easier review and merge.

PR #128, opened 2026-04-22
fix(qn-scale): handle 18-byte 0x12 frame from Renpho ES-26M

fromport:
### Problem

The Renpho ES-26M (Elis 1) sends an **18-byte** scale-info `0x12` frame on FFF1, where:
- byte[1] = packet length (0x12 = 18)  
- bytes[2-7] = device MAC address

The existing code assumed the classic ~11-byte format and read byte[2] as the protocol type. On the ES-26M this yielded `proto=0xFF` (actually a MAC address byte), causing all subsequent handshake commands (`0x13` config, `0x20` time sync, `0x22` start measurement) to be silently rejected by the scale. The full handshake completed without errors, but **no `0x10` weight frames were ever returned**.

### Fix

**1. Detect the long 0x12 frame format** (`data.length >= 18 && data[1] === data.length`):
   - Use `proto=0x00` instead of reading the MAC byte as protocol type
   - Set `weightScaleFactor=10` (ES-30M format); the existing heuristic fallback auto-corrects to `/100`

**2. Accept stable frames with R1=R2=0 in ES-30M mode:**
   - The ES-26M sends impedance=0 when the user wears socks (no skin contact with electrodes)
   - Previously these frames were unconditionally skipped, meaning users with socks never got a reading
   - When barefoot, the scale sends a final stable frame with R1/R2 > 0, which `isComplete()` accepts via the `impedance > 200` check

### Backwards compatibility

- Classic QN scales (11-byte `0x12` frame) are unaffected — the detection is length-based
- ES-30M scales with impedance still work — the `R1=R2=0` skip was only relevant when an impedance frame follows, and `isComplete()` already handles this via the `impedance > 200` gate

### Testing

Tested on:
- **Scale:** Renpho ES-26M (BLE name: `Renpho-Scale`, MAC: `FF:04:00:14:AC:0F`)
- **Services:** FFF0 (FFF1 notify, FFF2 write) + AE00 (AE01 write, AE02 notify)
- **Platform:** Debian (Asahi Linux) on Apple M2 Mac Mini, BlueZ D-Bus (node-ble handler)
- **Results:** 97.85 kg / 501 Ohm — both weight-only (socks) and full BIA (barefoot) readings confirmed working

### DIFF
--- a/src/scales/qn-scale.ts
+++ b/src/scales/qn-scale.ts
@@ -11,7 +11,8 @@ import { uuid16 } from './body-comp-helpers.js';
 import { bleLog } from '../ble/types.js';
 
 /** Format bytes as hex string for debug logging. */
-const hex = (data: number[]): string => data.map((b) => b.toString(16).padStart(2, '0')).join(' ');
+const hex = (data: number[] | Buffer): string =>
+  [...data].map((b) => b.toString(16).padStart(2, '0')).join(' ');
 
 /**
  * Ported from openScale's QNHandler.kt
@@ -47,9 +48,15 @@ const hex = (data: number[]): string => data.map((b) => b.toString(16).padStart(
  *   [7-8]   resistance R1 (BE uint16)
  *   [9-10]  resistance R2 (BE uint16)
  *
- * 0x12 frame (scale info):
+ * 0x12 frame (scale info, classic 11-byte format):
  *   [2]     protocol type (echoed back in all config commands)
  *   [10]    weight scale flag (1 = /100, else /10)
+ *
+ * 0x12 frame (ES-26M long format, 18 bytes):
+ *   [1]     length (== packet length, i.e. 0x12 == 18)
+ *   [2-7]   MAC address (NOT protocol type!)
+ *   Protocol type should be set to 0x00 for this variant.
+ *   Weight scale factor is 10 (ES-30M format with heuristic /100 fallback).
  */
 
 // Type 2 UUIDs (most common variant)
@@ -70,6 +77,14 @@ const SVC_T2 = 'fff0';
 
 /** Seconds from Unix epoch to 2000-01-01 00:00:00 UTC. */
 const SCALE_EPOCH_OFFSET = 946684800;
+
+/**
+ * Grace period (ms) to wait for an impedance frame after the first stable
+ * R1=R2=0 frame on long-frame variants (e.g. ES-26M). If an impedance frame
+ * arrives within this window, it supersedes the weight-only reading. If not,
+ * the weight-only reading is accepted on the next stable frame.
+ */
+const IMPEDANCE_GRACE_MS = 1500;
 
 export class QnScaleAdapter implements ScaleAdapter {
   readonly name = 'QN Scale';
@@ -95,6 +110,22 @@ export class QnScaleAdapter implements ScaleAdapter {
 
   /** Whether the AE00 service is available (newer firmware). */
   private hasAe00 = false;
+
+  /**
+   * Whether the scale sent a long-frame (18-byte) 0x12 variant (e.g. ES-26M).
+   * These scales may never provide impedance, so stable frames with R1=R2=0
+   * must be accepted after a grace period. Classic ES-30M scales always send
+   * an impedance frame after the weight-only stable frame, so skipping
+   * R1=R2=0 is correct there.
+   */
+  private isLongFrameVariant = false;
+
+  /**
+   * Timestamp (Date.now()) of the first stable R1=R2=0 frame seen on a
+   * long-frame variant. After IMPEDANCE_GRACE_MS without an impedance frame,
+   * subsequent R1=R2=0 stable frames are accepted.
+   */
+  private firstStableNoImpedanceAt: number | null = null;
 
   /** Deduplication guards: prevent duplicate state machine responses. */
   private configSent = false;
@@ -146,6 +177,8 @@ export class QnScaleAdapter implements ScaleAdapter {
     this.seenProtocolType = 0x00;
     this.weightScaleFactor = 100;
     this.hasAe00 = false;
+    this.isLongFrameVariant = false;
+    this.firstStableNoImpedanceAt = null;
     this.configSent = false;
     this.timeSyncSent = false;
     this.historyResponseSent = false;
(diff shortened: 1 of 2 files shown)


KristianP26:
Thanks for the thorough investigation and the field testing on ES-26M. The long-frame detection (fix 1) looks good and is nicely backwards compatible.

I have one concern about the second change that I think should be addressed before merging, plus a few smaller items.

**Regression risk from removing the `R1=R2=0` skip**

`isComplete()` at `src/scales/qn-scale.ts:502` accepts `impedance=0` when weight is positive:

```ts
if (reading.impedance === 0) return reading.weight > 0;
return reading.weight > 10 && reading.impedance > 200;
```

The skip at line 334 was added specifically to prevent the first ES-30M stable frame (weight only, no impedance) from completing before the impedance frame arrives. With the skip removed, a barefoot ES-30M user will complete on the weight-only frame and lose the BIA reading.

This is also covered by an existing test at `tests/scales/qn-scale.test.ts:439`:

```ts
// Stable frame (state=0x02) but R1=R2=0 (impedance not measured yet)
...
// Should return null because impedance isn't ready yet
expect(adapter.parseNotification(buf)).toBeNull();
```

Your comment mentions a heuristic ("if the scale sends multiple stable frames with R1=R2=0, accept them"), but the code does not implement one. Could we gate the skip on a long-frame marker set in the 0x12 handler so ES-30M keeps its current behavior and ES-26M gets the new one? Something like:

```ts
// in the 0x12 handler
if (data.length >= 18 && data[1] === data.length) {
  this.isLongFrameVariant = true;
  this.seenProtocolType = 0x00;
  this.weightScaleFactor = 10;
}
```

```ts
// in the ES-30M branch
if (!this.isLongFrameVariant && stable && r1 === 0 && r2 === 0) return null;
```

A counter or timer would also work, but a flag is simplest and keeps the existing ES-30M contract intact.

**Smaller items**

1. Please rebase onto `dev`. We merge feature PRs into `dev` and only release PRs into `main`.
2. A test for the 18-byte 0x12 frame would be great (sets `proto=0x00` and `factor=10`, and a classic 11-byte frame still reads `proto=data[2]`).
3. The `tests/scales/qn-scale.test.ts:439` test will need to be updated or kept depending on the chosen semantics.
4. Before pushing, `npm test && npm run lint && npx tsc --noEmit` should all pass.

Again, really appreciate the deep protocol analysis and the clean explanation in the PR description. Once fix 2 is gated, this will close out a long-standing issue for ES-26M owners.

KristianP26:
One more small thing: could you run `npm run format` before pushing? The new inline debug line at the top of `parseNotification` likely exceeds the 100-char Prettier width, and a clean format pass will keep the diff minimal.

fromport:
Updated per review feedback:

Added isLongFrameVariant flag to gate the R1=R2=0 skip (preserves existing ES-30M behavior)
Rebased onto dev
Extracted hexBuf() helper to keep debug line within Prettier width

Note: I couldn't retarget the base branch from main to dev via the GitHub UI. Could you retarget it, or shall I close this and open a new PR against dev?
Still TODO: tests for the 18-byte 0x12 frame and updated ES-30M R1=R2=0 test semantics.

KristianP26:
Thanks for the fix, problem breakdown is clear and the fallback heuristic makes sense.

Two things I'd like to see before merge:

**1. Missing tests**

`tests/scales/qn-scale.test.ts` covers classic + ES-30M. Please add:
- 18B `0x12` frame sets `isLongFrameVariant=true`, `proto=0x00`, `factor=10`
- long-frame variant: stable `R1=R2=0` 0x10 frame returns as reading (socks path)
- classic ES-30M: stable `R1=R2=0` is still skipped (regression guard)
- long-frame barefoot: `R1>0` returns impedance correctly

**2. Potential early-completion race**

Long-frame branch returns `{weight, impedance: 0}`. `isComplete()` has `if (reading.impedance === 0) return reading.weight > 0`, which is immediately `true`, and the orchestrator in `src/ble/shared.ts:262` stops reading.

If ES-26M (or a future firmware revision) emits a transient stable R1=R2=0 frame **before** the impedance-bearing one (the way classic ES-30M does), BIA would be lost even in the barefoot case. It worked on your hardware, but without a fallback timer or "wait for second stable" guard this is fragile.

Possible mitigation: accept an R1=R2=0 reading only after a short delay (e.g. 500 ms) without an impedance frame, or after a second stable frame with the same weight.

**Minor**

- `hexBuf` is used once, could fold into `hex()` via overload (not a blocker)
- Detection `data[1] === data.length` is a narrow heuristic but safe (classic 0x12 is ~11 B)

Otherwise nice protocol analysis, scoped PR, classic + ES-30M backwards compat looks preserved.

fromport:
Updated per review:

Added isLongFrameVariant flag to gate the R1=R2=0 skip (preserves ES-30M behavior)
Added impedance grace period (1500ms) to prevent early-completion race: first stable R1=R2=0 frame is skipped, accepted only after grace period elapses without an impedance frame
Folded hexBuf into hex() overload
Added 7 new tests covering: 18B 0x12 frame detection, classic 11B backwards compat, barefoot impedance, grace period skip/accept/supersede, and ES-30M regression guard
npx vitest run: all scale tests pass (2 pre-existing failures in wizard/mqtt tests due to missing dev deps, unrelated to this PR)
tsc --noEmit: no errors in qn-scale.ts
prettier --write: no changes needed

KristianP26:
Thanks a lot for the thorough investigation and clean fix! 🙌

The root cause write-up is excellent — figuring out that byte[2] was a MAC byte masquerading as a protocol type (and that every subsequent command was silently rejected because of it) is exactly the kind of subtle handshake bug that's painful to track down without a real device in hand. Really appreciate you including the captured raw frames in both the description and the tests, it makes the fix trivial to verify.

A few things I especially liked:

- Length-based detection (`data.length >= 18 && data[1] === data.length`) keeps classic QN scales untouched — no risk of regression for existing users.
- The grace-period path for R1=R2=0 is a nice touch. Handling the socks-vs-barefoot case gracefully means users get a reading either way, and the impedance frame still wins when it arrives.
- 7 new tests covering both the long-frame path and a regression guard for the classic ES-30M skip. Exactly the kind of coverage I'd have asked for.

Merged to `dev`. Will ship in the next release (likely v1.10.2 or v1.11.0 in the coming days, bundled with a couple of other in-flight items). I'll ping this thread once it's out so you can drop the patch.

Thanks again for contributing — this is the second Renpho/QN variant we've nailed down thanks to community BLE captures, and it genuinely makes the project better. 🙏

Issue #134, opened 2026-04-23
Support Xiaomi Mi Body Composition Scale 2 (XMTZC05HM)


pr1moz:
### Description

Would love to see support added for the Xiaomi Mi Body Composition Scale 2 (XMTZC05HM / Mi Smart Scale 2). This scale is not currently supported by this project (I tried before checking), and as far as I can see, it is also not supported by OpenScale, which is mentioned as one of the main sources for adapters.

From my research, the scale uses a different BLE advertisement format than the older Xiaomi scales. It advertises as `MIBFS`, uses BT 5.0 LE, and broadcasts weight + impedance passively in advertisement packets, no pairing needed. The service UUID is `0x181B` and the payload structure is different from the gen 1 scale.

The protocol is already well documented and implemented in several other projects that could be used as reference:

- **ESPHome** `xiaomi_miscale2` - explicitly supports XMTZC05HM, decodes weight and impedance from passive BLE ads: https://esphome.io/components/sensor/xiaomi_miscale/
- **lolouk44/xiaomi_mi_scale** - Python lib on PyPI that handles this packet format: https://github.com/lolouk44/xiaomi_mi_scale
- **zibous/ha-miscale2** - Python/MQTT bridge built specifically for this model: https://github.com/zibous/ha-miscale2
- **Theengs Decoder** - has a device profile for XMTZC05HM: https://decoder.theengs.io/devices/XMTZC05HM.html

### Use Case

I have this scale and want to sync weight + body composition to Garmin Connect through Home Assistant automatically, instead of having to open an app on my phone every time. That's exactly what this project does, just not for this scale yet :)

KristianP26:
@pr1moz thanks for the well-researched request. The Mi Body Composition Scale 2 (XMTZC05HM / MIBFS) is a great candidate. The protocol is fully public, it is a pure broadcast scale (no connection, no pairing), and the references you linked (ESPHome `xiaomi_miscale2`, Theengs Decoder, lolouk44, zibous) all agree on the packet format. That makes this straightforward to add rather than a reverse-engineering project.

Quick summary of what the adapter needs to do, for anyone following along:

- Listen for advertisement packets with service data UUID `0x181B` (the Xiaomi Body Composition service).
[...]

PR #137, opened 2026-04-25
Feature/Mi Body Composition Scale 2

boildead:
## Summary

The Xiaomi Mi Scale 2 (Mi Body Composition Scale 2/ MIBFS / XMTZC05HM) connects via GATT but never streams notifications on the history characteristic — a known firmware variant behaviour. ESPHome's `xiaomi_miscale2` component and Theengs Decoder both work around this by decoding the same 13-byte frame directly from the BLE advertisement's service data (UUID 0x181B), with no pairing or connection required.

This PR adds a passive broadcast decode path for the Mi Scale 2 that works the same way, and extends all four BLE transport handlers to dispatch service-data advertisements so the pattern is available to future adapters too.

## Changes

- **`ScaleAdapter` interface** — added optional `parseServiceData(uuid, data)` method and `preferPassive` flag; added `serviceData` array to `BleDeviceInfo`
- **`MiScale2Adapter`** — implemented `parseServiceData` for UUID 0x181B (same 13-byte frame already used in GATT path); set `preferPassive = true` to skip the GATT connect entirely; extended `matches()` to fall back to service-data UUID when local name is absent (ESPHome omits it)
- **All four BLE handlers** (node-ble, noble, noble-legacy, esphome-proxy, mqtt-proxy) — extract service data from advertisements and call `parseServiceData` alongside the existing `parseBroadcast` path
- **Grace period** (`IMPEDANCE_GRACE_MS = 12 s`) — the scale broadcasts weight-only frames while BIA is in progress, then a final frame with impedance once complete (~10–20 s). Handlers now wait up to 12 s for the impedance-bearing frame before falling back to weight-only
- **node-ble PropertiesChanged fix** — the D-Bus subscription was using the wrong field name; simplified to `helper.on('PropertiesChanged', ...)` via node-ble's built-in BusHelper event emitter
- **ESPHome GATT warning fix** — warning for GATT-only scales now also fires for dual-mode adapters whose broadcast path returns null (e.g. Elis 1 MAC beacons)
- **Docs** — updated `body-composition.md` with Mi Scale 2 algorithm notes; updated `esphome-proxy.md` with scan duty-cycle tip and GATT section; updated supported scales table

## Test Plan

- [x] Tests pass (`npm test`)
  ```
  Test Files  64 passed (64)
        Tests  1261 passed (1261)
  ```

Tested on a real XMTZC05HM with:
- **Direct BLE (node-ble / Linux)** — passive mode resolves without any GATT connection; impedance-bearing frame captured via `PropertiesChanged` subscription; weight-only fallback fires after 12 s when impedance frame is missed
- **ESPHome proxy** — scale correctly matched via service-data UUID (not GATT); scan duty-cycle tip in docs confirmed to eliminate "second weigh-in" issue
- **ESP32 BLE Proxy** — I don't have compatible hardware so I wasn't able to test it

## Notes

The GATT unlock path (`ENABLE_HISTORY_MAGIC` command) is intentionally preserved as a fallback for firmware variants that do stream on the characteristic. `preferPassive = true` routes the scale through broadcast scan first; GATT is only attempted if the handler does not support service-data dispatch (none of the current handlers fall into that category).

The 12-second grace period is a per-weigh-in timer — it starts on the first stable weight-only frame and resets between readings. It has no effect on scales that already include impedance in every frame.


KristianP26:
Thanks @boildead, this is a substantial and well-scoped contribution. Tests, lint and tsc all clean here too. A few things before I merge:

## Main concern — `isComplete` gate widened to all broadcast adapters

The old `ReadingWatcher` flow in `handler-esphome-proxy.ts` and `handler-mqtt-proxy.ts` accepted any `parseBroadcast` reading immediately. The new flow gates on `adapter.isComplete(reading)` and falls back to a 12s grace timer otherwise. Same change in `broadcastScan` of `handler-noble.ts` / `handler-noble-legacy.ts`.

That gate is correct for Mi Scale 2 (weight-only frame must wait for impedance), but it now applies to every existing broadcast adapter. Any adapter whose `parseBroadcast` returns a reading where `isComplete` is false will get a 12s latency penalty it did not have before.

Could you either:
1. Scope the grace path to `preferPassive` adapters only (cleanest, matches the feature intent), or
2. Audit the broadcast-capable adapters and confirm their `isComplete` returns true for stable broadcast frames.

## Smaller items

**`.gitignore`** — `.claude/` belongs in `.git/info/exclude` (local) per project convention, not the tracked file. Also missing trailing newline. Please drop this hunk.

**Cross-address mix in `handler-esphome-proxy.ts` `scanAndReadRaw`** — `graceTimer` and `bestWeightOnly` are scan-scoped instead of per-address. If two scales broadcast partial frames in the same scan, the log line uses the first address but the reading is from the latest. `ReadingWatcher` already keys by address; same pattern here would be consistent. Edge case, optional.

**`handler-mqtt-proxy.ts` fallback** — `if (reading && !weightOnlyFallback)` keeps the first partial; esphome path keeps the latest. Pick one. `registerScaleMac(config, candidates[0]?.address ?? '')` can pass the wrong MAC if the fallback came from a non-first candidate. Capture the address alongside the reading.

**`mi-scale-2.ts` UUID normalization** — `parseServiceData` only expands the 4-char short form. Current handlers normalize upstream so this works, but the 8-char unpadded form (`0000181b`) would fall through. Trivial.

**Test coverage** — no tests for the grace-timer logic in any handler. With four handlers and three branches each (complete / partial-then-complete / partial-then-timeout) it would help to lock the behavior in. Vitest fake timers work well here.

**Lint** — 8 `eslint-disable @typescript-eslint/no-explicit-any` lines in `broadcastScanNodeBle`. The D-Bus surface is genuinely loose, but a single local `type DbusValue = ...` plus one helper would cut most of them.

## Things I liked
- Clean interface extension, fully backward-compatible
- node-ble passive path is solid: `DuplicateData=true` filter + `PropertiesChanged` BusHelper event + 500ms poll fallback, with `extractDbusBytes` covering all the variant shapes
- `logPhase1Capabilities` correctly counts `parseServiceData`-only adapters as broadcast-capable
- Good `parseServiceData` test matrix (short / dashed / normalized UUIDs) and `matches()` coverage for both serviceUuids and serviceData
- Grace timer cleanup wired through both watchers' teardown paths

boildead:
Hi @KristianP26, thanks for review. I'll have more time on weekend to address those.

KristianP26:
@
@boildead no rush, weekend pass works for me. Posting a quick prioritization so you don't have to guess what's blocking vs. optional:

**Must-fix to merge:**
1. `isComplete` grace-timer scoped to `preferPassive` adapters only (option 1 from my review). Main correctness concern — a 12 s latency penalty for every existing broadcast adapter is the part I can't ship.
2. CI `typecheck-and-test (Node 20)` is failing on `62c12dc`. Please pull the run log and address — the 22/24 jobs are cancelled but I assume they would fail the same way.

**Would prefer in this PR but won't block:**
- Drop the `.gitignore` `.claude/` hunk
- mqtt-proxy: capture address alongside the partial reading so `registerScaleMac` can't bind the wrong MAC
- `parseServiceData` 8-char unpadded UUID (`0000181b`) handling

**Fine as a follow-up issue:**
- Per-address grace state in esphome-proxy `scanAndReadRaw`
- Lint cleanup of the 8 `any`s in `broadcastScanNodeBle`
- Vitest fake-timer tests for the 3 grace-timer branches per handler — happy to open this as a separate issue and tackle it myself after merge if you'd rather not.

Ping me when you push and I'll re-review same day.
@

KristianP26:
@boildead I went ahead and pushed `045665b` to take care of the must-fix items so we can land this for #134. You're credited as `Co-Authored-By` on the commit. Quick summary:

**Must-fix #1 — preferPassive scope (resolved)**

`isComplete` gate + 12s grace timer are now scoped to `adapter.preferPassive === true` in all four broadcast paths (`handler-esphome-proxy` × 2, `handler-mqtt-proxy` × 2, `handler-noble`, `handler-noble-legacy`). Existing broadcast adapters (Eufy P2, QN-Scale) emit any non-null reading immediately, exactly as before this PR. Mi Scale 2 keeps the new behavior because of `preferPassive = true` on the adapter.

**Must-fix #2 — CI Node 20 (resolved)**

The failure was Prettier `format:check`, not typecheck. Six files had drift from prior commits (`handler-node-ble.ts` was the last one not yet covered); fixed in place.

**Smaller items addressed:**
- mqtt-proxy `weightOnlyFallback` now captures the source address alongside the reading, so `registerScaleMac` cannot bind to a non-first candidate's MAC.
- mi-scale-2 `parseServiceData` accepts the 8-char unpadded UUID form (`0000181b`) in addition to the 4-char short form.
- Dropped the `.gitignore` `.claude/` hunk; restored trailing newline.

**Still pending (left for you / follow-up if you want):**
- Cross-address mix in `handler-esphome-proxy` `scanAndReadRaw` (per-scan rather than per-address `graceTimer`/`bestWeightOnly`). Edge case.
- Lint cleanup (8 `no-explicit-any` disables in `broadcastScanNodeBle` could fold into a `DbusValue` helper).
- Tests for the grace-timer logic in handlers (vitest fake timers).

Local quality gates all green: `tsc`, `eslint`, `prettier --check`, 1275/1275 tests.

Thanks again for the substantial work on this — the passive scan plumbing, `parseServiceData` interface, and the `extractDbusBytes` D-Bus value normalization are all clean. Once CI is green I'll re-target the base from `main` to `dev` per the project merge convention.

PR #145, opened 2026-04-27
DRAFT: #139 Add BLE Proxy support for ESP-WROOM-32

APIUM:
## Summary

Adds support for BLE Proxy ESP-WROOM-32 -> https://github.com/KristianP26/ble-scale-sync/issues/139

This is in draft until tested by @geniusliang

### DIFF
--- a/PORTING.md
+++ b/PORTING.md
@@ -2,7 +2,8 @@
 
 ## Adding a headless board (no display) — ~5 minutes
 
-1. Copy `firmware/board_esp32_s3.py` to `firmware/board_<name>.py`
+1. Copy `firmware/board_esp32_s3.py` (or `board_esp_wroom_32.py` for a stock
+   ESP32 with software-shared BLE/WiFi radio) to `firmware/board_<name>.py`
 2. Edit constants: BOARD_NAME, radio coexistence, scan timing, GC, beep
 3. Add dispatch entry in `firmware/board.py`
 4. Add flash config in `firmware/flash.sh` (chip, firmware URL, baud, offset)
--- a/docs/guide/esp32-proxy.md
+++ b/docs/guide/esp32-proxy.md
@@ -49,10 +49,11 @@ Any ESP32 board running MicroPython with BLE support works. Tested on:
 | Board                          | Price | Notes                                                                 |
 | ------------------------------ | ----- | --------------------------------------------------------------------- |
 | M5Stack Atom Echo (ESP32-PICO) | ~8€   | Tiny, no PSRAM, ~100 KB free RAM, I2S buzzer for beep feedback        |
+| Generic ESP-WROOM-32 dev board | ~5€   | Stock ESP32 module, no display, BLE/WiFi share radio (slower scans)   |
 | ESP32-S3-DevKitC               | ~12€  | Standard dev board, plenty of RAM                                     |
 | Guition ESP32-S3-4848S040      | ~25€  | 480x480 RGB display, shows scan status and export results via LVGL UI |
 
-The board is auto-detected from the chip family. Set `"board"` in `config.json` to override (e.g. `"guition_4848"` for the display board, `"atom_echo"` for the Atom Echo).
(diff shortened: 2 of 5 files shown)


KristianP26:
Thanks @APIUM, looks good as a draft. The shape matches `board_atom_echo.py` cleanly and `flash.sh` already exports `MICROPYTHON_DATE_ESP32`, so the URL builds correctly.

A few items to address before un-drafting:

**Required**

1. Replace the em dash in the new file's comment with a plain hyphen. Project style avoids em dashes in code and docs:
   ```python
   # BLE/WiFi coexistence - shared radio, must deactivate BLE after scan
   ```
   (`board_atom_echo.py` has the same em dash, that one is a pre-existing inconsistency, no need to touch it in this PR.)

2. Expand the PR description with a short test plan for @geniusliang to tick off, so we know what "tested" means here:
   - [ ] `./flash.sh --board esp_wroom_32` completes without errors on a stock WROOM-32
   - [ ] Board boots, joins WiFi, MQTT broker connects
   - [ ] Scale is discovered and a measurement reaches the main app via the proxy
   - [ ] WiFi reconnects cleanly after each scan cycle (the whole point of `DEACTIVATE_BLE_AFTER_SCAN=True`)
   - [ ] At least 3 successful back-to-back measurements

**Nice to have**

3. One-line note in `README.md` if you want, listing WROOM-32 next to the other supported boards. Optional, README is brief.

**Out of scope, follow-up only**

The auto-detect in `firmware/board.py` still routes any non-S3 ESP32 to the `atom_echo` profile, which means a WROOM-32 user who forgets the `config.json` override (or the `--board` flag on first flash) ends up with `HAS_BEEP=True` and I2S init on pins 19/33/22. Probably harmless on a bare module, but a real risk if the user wired those pins to their own hardware. The docs note and the updated `flash.sh` warning mitigate it for now. I'll open a follow-up to make the auto-detect fallback a generic no-peripherals profile, with `atom_echo` requiring an explicit override. Not blocking this PR.

Once the test plan is checked and the em dash swapped, flip out of draft and I'll merge.

KristianP26:
Merged into `dev` (resolving the conflict) and followed up with a memory-hardening fix, since flashing alone wasn't enough — the WROOM-32 was stuck in a reboot loop.

**Conflict resolution (`firmware/flash.sh`):** kept `dev`'s pinned async-primitives ref `github:peterhinch/micropython-async/v3/primitives@68b5f01e999b` (the canonical #199 fix) over the PR's unpinned install, for reproducibility. The PR's `MICROPYTHON_DATE_ESP32=20251209` was kept — verified the binary is live, and it also corrects a latent stale date on the Atom Echo path. All other files merged cleanly.

**Root cause of @geniusliang's instability** (from the serial dumps): `board_esp_wroom_32.py` assumed ~250 KB free RAM, but a stock module has only ~90 KB free after `aioble` + `mqtt_as` imports (same ballpark as the Atom Echo). With `MAX_SCAN_ENTRIES = 300` in a noisy BLE environment the raw scan buffer exhausted the heap mid-scan; the IRQ `append` raised `MemoryError`, the unhandled IRQ exception left NimBLE in a bad state, and the device panicked on a failed semaphore allocation (`npl_freertos_sem_init` → Guru Meditation → reboot loop, status flapping online/offline).

**Hardening commit (`c99605e`):**
- `board_esp_wroom_32.py`: corrected the RAM docstring, `MAX_SCAN_ENTRIES` 300 → 80, `SCAN_DURATION_MS` 8000 → 5000, em dash → hyphen (per the earlier review note).
- `ble_bridge.py`: wrapped the IRQ append in `try/except MemoryError` in **both** the batch `scan()` and streaming IRQ handlers, so a transient OOM drops the advertisement instead of crashing the BLE stack. This also hardens the Atom Echo.

**Note for tuning:** `MAX_SCAN_ENTRIES = 80` is a conservative starting point, not an empirically tuned value — the IRQ guard is what actually stops the hard crash. @geniusliang, please re-flash from `dev` and report: if the scale is reliably picked up it's good as-is; if it's missed in your environment we can raise the cap (the guard makes that safe). Thanks for the very detailed logs — they made this diagnosable.

The auto-detect fallback (non-S3 → `atom_echo`) is tracked separately as a follow-up.

PR #227, opened 2026-06-04
feat(scales): support Smart Scale A1 broadcasts

peipeipe:
## Summary
- enable the 1byone/Eufy adapter to prefer passive BLE advertisement reads
- parse Smart Scale A1-style ManufacturerData broadcasts by locating embedded 0xCF measurement frames
- let the node-ble broadcast scanner handle ManufacturerData as well as ServiceData
- keep retry cadence short for scales that power off quickly
- ignore local Garmin token and systemd service files so secrets and machine-specific config are not committed

## Tests
- npm run build
- npm test

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -39,6 +39,10 @@ firmware/*.bin
 # Garmin tokens (should not be committed)
 .garmin_tokens/
 .garmin_renpho_tokens/
+garmin-tokens/
+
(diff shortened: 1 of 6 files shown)


peipeipe:
Closing for now to narrow the scope before resubmitting.

Issue #213, opened 2026-05-28
App will not discover scale after running for a week


mikeeales:
### Description

Docker App will not discover scale after running for a week

### Steps to Reproduce

Run app for a long period

### Scale Brand / Model

Renpho

### Platform

Docker

### Node.js Version

20.11.0

### App Version

1.14.1

### Debug Logs

```shell
2026-05-27 05:50:47.557 [Sync] Raw reading: 83.15 kg / 508 Ohm
2026-05-27 05:50:47.558 [Sync] [Mike] Matched (tier: exact)
2026-05-27 05:50:47.686 [Sync] [Mike] Measurement: 83.15 kg / 508 Ohm
2026-05-27 05:50:47.686 [Sync] [Mike] Body composition:
2026-05-27 05:50:47.687 [Sync] [Mike]   bmi: 26.24
2026-05-27 05:50:47.687 [Sync] [Mike]   bodyFatPercent: 34.43
2026-05-27 05:50:47.687 [Sync] [Mike]   waterPercent: 47.86
2026-05-27 05:50:47.687 [Sync] [Mike]   boneMass: 2.29 kg
2026-05-27 05:50:47.688 [Sync] [Mike]   muscleMass: 29.44 kg
2026-05-27 05:50:47.688 [Sync] [Mike]   visceralFat: 19
2026-05-27 05:50:47.688 [Sync] [Mike]   physiqueRating: 1
2026-05-27 05:50:47.688 [Sync] [Mike]   bmr: 1684
2026-05-27 05:50:47.688 [Sync] [Mike]   metabolicAge: 62
2026-05-27 05:50:47.689 [Sync] Exporting to: garmin, strava...
2026-05-27 05:50:47.732 [Strava] Access token expired, refreshing...
2026-05-27 05:50:47.914 [Sync] Update available: v1.15.0 (current: v1.14.1). See https://blescalesync.dev/changelog
2026-05-27 05:50:48.026 [Strava] Token refreshed successfully.
2026-05-27 05:50:48.283 [Strava] Strava weight updated to 83.15 kg.
[Garmin] Loading tokens from garmin-tokens
[Garmin] Authenticated.
[Garmin] Uploading body composition...
[Garmin] Upload successful!
2026-05-27 05:50:50.678 [Garmin] Garmin upload succeeded.
2026-05-27 05:50:50.678 [Sync] Done.

2026-05-27 05:50:50.679 [Sync] Waiting 30s before next scan...
[...]

PR #233, opened 2026-06-11
fix(qn-scale): support store-and-forward variants that deliver 0x23 records

built-dev:
## Problem

Some QN-protocol Renpho units never stream live `0x10` weight frames. Confirmed on a real **Renpho Elis 1** (advertises as `Renpho-Scale`, FFF0 + AE00 services, 18-byte `0x12` scale info with `factor=10`): the scale measures autonomously, stores the finished measurement, and delivers it as a `0x23` record right after the handshake completes. Since the adapter ignores `0x23` ("historical record, no action needed"), every session ends in `Scale disconnected before reading completed` and the scale produces **no readings at all**.

There's a second wrinkle: an always-on listener connects the moment the scale wakes (it only advertises when stepped on), i.e. **mid-measurement** — at which point the scale reports an *empty* history and never volunteers the fresh record on its own.

## Observed 0x23 layout (19 bytes, captured from hardware)

| bytes | meaning |
|---|---|
| [3] | total record count (`00` = empty "no records" frame) |
| [4] | record index, 1-based |
| [6-9] | measurement timestamp, LE, seconds since 2000-01-01 |
| [10-11] | weight, BE, factor-scaled (existing alternate-factor heuristic applies) |
| [12-13] / [14-15] | R1 / R2 BIA resistances |

Example real frame: `23 13 ff 01 01 f0 aa c9 bd 31 1e 5a 01 f1 01 ef 00 00 e2` → 77.7 kg, R1=497.

## Fix

1. **Parse `0x23` records** as readings, deduplicating replays across reconnects via the record's scale-side timestamp.
2. **Nudge for the fresh record**: when the history is empty (connected mid-measurement), re-send the `0x22` start command every 5 s (max 12×). The scale then delivers the new record in-session — typically by the third nudge, ~15–20 s after step-on. Timer is cleaned up on reconnect/record/write-failure and never starts without an active connection.

Live `0x10` parsing is untouched; scales that stream normally behave exactly as before (`0x23` empty frames merely arm the nudge, and a delivered `0x10` reading was already terminal).

## Testing

- 6 new Vitest cases for the `0x23` path (real-capture frame, factor heuristic, multi-record session, empty terminator, replay dedup, R2 fallback) in the existing `qn-scale.test.ts` style.
- Full suite: **1630/1630 passing**, `tsc --noEmit` and `eslint` clean.
- Hardware-verified end-to-end (Linux/BlueZ via node-ble, continuous mode): previously zero readings; with this change every weigh-in lands whether the connection happens before, during, or after the measurement.

This may explain other "connects fine but 'disconnected before reading completed'" reports on Renpho/QN units.

### DIFF
--- a/src/scales/qn-scale.ts
+++ b/src/scales/qn-scale.ts
@@ -133,6 +133,13 @@ export class QnScaleAdapter implements ScaleAdapter {
    */
   private firstStableNoImpedanceAt: number | null = null;
 
+  /**
+   * Scale-side timestamps (seconds since 2000) of accepted 0x23
+   * stored-measurement records. Guards against the scale replaying the same
+   * record on reconnect. Deliberately NOT reset between connections.
+   */
+  private seenHistoryTimestamps = new Set<number>();
+
   /** Deduplication guards: prevent duplicate state machine responses. */
   private configSent = false;
   private timeSyncSent = false;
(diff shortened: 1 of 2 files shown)


KristianP26:
Closing this.

The 0x23 stored-record support it adds is already implemented and shipped: `5617f28` ("parse QN 0x23 stored record for V10 Renpho/ES-CS20M", #213) has been on `dev` and in the v1.17.0 release for a while, and it was validated against real hardware by the reporter. This PR branched before that landed, so it is a parallel duplicate and now conflicts with `dev`. CI is also red on every Node version.

A few things to keep in mind for any future PR here:

- **It must be rebased and green.** This one conflicts with `dev` and fails `typecheck-and-test`. PRs in that state cannot be reviewed or merged.
- **Tests must prove something real.** The added tests build a synthetic 0x23 frame with one set of byte offsets and then parse it back with the same offsets, so they pass no matter what the real frame layout is. They are described as "captured from real hardware", but a self-referential test does not demonstrate that. For a protocol change I need the actual HCI capture (`btsnoop_hci.log` / `.pklg`) plus the ground-truth weight and impedance the official app showed, so the byte mapping can be checked against reality. Notably this PR reads the resistances at different offsets/endianness than the shipped code, with nothing to confirm which is correct.
- **Please do not open unverified, machine-generated adapter changes.** Decoded protocol work has to be grounded in a real capture and a real measurement, otherwise it cannot be trusted or maintained.

If you have a QN variant that still does not read on v1.18.0, please open a fresh issue with an HCI capture and the ground-truth values and I will look into it.

Issue #117, opened 2026-04-20
Support for the R-MSC04 Scale


armontini:
### Description

Renpho released the R-MSC04 Body Composition Scale, which provides a more in depth analysis in comparison to the Elis 1. Could this ble scale be added?

### Use Case

This scale provides the same 10 pieces of information, while adding additional information such as individual limb statistics, subcutaneous fat, protein percentage, and more. Refer to the official website for me info: https://renpho.com/products/morphoscan-nova-body-composition-analyzer?variant=44798210572460

### Alternatives Considered

Tried to add this before opening issue, but the add-on kept failing to find/add the scale.

KristianP26:
@armontini thanks for the request. The R-MSC04 (MorphoScan Nova) looks like a different protocol family from the Renpho models we support today. Our existing Renpho/QN-Scale adapter covers the older Elis 1 / ES-CS20M / Senssun / Sencor devices over the classic FFE0/FFF0 GATT services. The MSC04 is an 8-electrode handle scale with a TFT display, which usually means different services and possibly a different advertising format. openScale also does not have an issue or adapter for this model yet.

Since scanning fails to find the scale on your end, the very first thing to pin down is whether the scale is advertising at all when our scanner looks, and if so, what name and services it exposes.

To build an adapter I need three pieces of data:

### 1. Raw scan output

Run the scanner from the add-on or Docker image and share everything the scale advertises:

```bash
docker run --rm -it --network host --cap-add NET_ADMIN --cap-add NET_RAW \
  --group-add 112 -v /var/run/dbus:/var/run/dbus:ro \
  -e DEBUG=true \
  ghcr.io/kristianp26/ble-scale-sync:latest scan
```

Step on the scale (or press its pair button) while the scan runs, so we capture its broadcast. Please include everything, including entries that are not recognized.

### 2. Diagnostic output

If the scale does get discovered by scan, run diagnose against its MAC for the full service and characteristic dump:

```bash
docker run --rm -it --network host --cap-add NET_ADMIN --cap-add NET_RAW \
  --group-add 112 -v /var/run/dbus:/var/run/dbus:ro \
  -e DEBUG=true \
  ghcr.io/kristianp26/ble-scale-sync:latest diagnose <SCALE_MAC>
```

### 3. HCI snoop log from the Renpho Health app

The big one. The Renpho Health app successfully talks to the scale, so its packet capture reveals the entire protocol (pairing, handshake, the 50+ metrics frame format).

[...]

Issue #230, opened 2026-06-05
R-MSC02 (Renpho MorphoScan Base) - unlock command not working


rns-e:
### Description

Scale: Renpho MorphoScan Base (R-MSC02)
Platform: macOS (darwin x64)
Version: 1.16.0

The scale is discovered and connected successfully via UUID 
57c361d94af2bcb4e063b785279349a3. Adapter ES-CS20M matches.

However, the unlock command [55 aa 90 00 04 01 00 00 00 94] 
receives no response from the scale. No notifications are 
received even when stepping on the scale.

Services discovered:
- 0x1A10 with notify 0x2A10 and write 0x2A11 (matched by ES-CS20M)
- 0x1530 (DFU service?)

The scale works normally with the Renpho Health app v7.8.0 on Android.

Is the unlock command different for R-MSC02 vs ES-CS20M/ES-32MD?

### Steps to Reproduce

Steps to Reproduce:
1. Configure ble-scale-sync with scale MAC/UUID for R-MSC02
2. Run npm start
3. Step on the scale and wait for stable reading
4. Scale completes measurement and turns off display
5. No data received in ble-scale-sync - only repeated unlock commands sent

### Scale Brand / Model

Renpho MorphoScan Base (R-MSC02)

### Platform

macOS

### Node.js Version

v22.15.0

### App Version

Renpho Health app - 7.8.0

### Debug Logs

```shell
2026-06-05 07:22:03.684 [BLE:debug] Target device matched: R-MSC02 [57c361d94af2bcb4e063b785279349a3]
2026-06-05 07:22:04.078 [BLE] Matched adapter: ES-CS20M
2026-06-05 07:22:04.078 [BLE:debug] Services: [0000180a00001000800000805f9b34fb, 00001a1000001000800000805f9b34fb, 000015301212efde1523785feabcd123]
2026-06-05 07:22:04.078 [BLE:debug] Service 00001a1000001000800000805f9b34fb: 3 characteristic(s)
2026-06-05 07:22:04.078 [BLE:debug]   Char 00002a1100001000800000805f9b34fb props=[write]
2026-06-05 07:22:04.078 [BLE:debug]   Char 00002a1000001000800000805f9b34fb props=[notify]
2026-06-05 07:22:04.078 [BLE:debug]   Char 00002a1200001000800000805f9b34fb props=[indicate]
[...]

PR #256, opened 2026-06-26
feat(ble): add Renpho R-MSC04 scale adapter and tests

udochrist:
## Summary

Add Support for R-MSC04 based on discovery and Apple PacketLogger.
Code was generated using claude.ai and is not yet manally tested. all contained tests are working, however.
this PR is more to inform about the work done and potential usage of this to prevent duplicate work analyzing the files.

## Changes

Added file for scale + tests as well as registered the new scale

-

## Test Plan

- [X] Tests pass (`npm test`)
- [X] Lint clean (`npm run lint`)
- [X] TypeScript compiles (`npx tsc --noEmit`)
- [ ] Tested manually (describe below)

## Notes

As above:
Code was generated using claude.ai and is not yet manally tested. all contained tests are working, however.
this PR is more to inform about the work done and potential usage of this to prevent duplicate work analyzing the files.

### DIFF
--- a/src/scales/index.ts
+++ b/src/scales/index.ts
@@ -2,6 +2,7 @@ import type { ScaleAdapter } from '../interfaces/scale-adapter.js';
 import { QnScaleAdapter } from './qn-scale.js';
 import { RenphoScaleAdapter } from './renpho.js';
 import { RenphoEs26bbAdapter } from './renpho-es26bb.js';
+import { RenphoMsc04Adapter } from './renpho-msc04.js';
 import { MiScale2Adapter } from './mi-scale-2.js';
 import { XiaomiS800Adapter } from './xiaomi-s800.js';
 import { BeurerBf720Adapter } from './beurer-bf720.js';
@@ -39,6 +40,7 @@ export const adapters: ScaleAdapter[] = [
   new QnScaleAdapter(),
   new RenphoScaleAdapter(),
   new RenphoEs26bbAdapter(),
+  new RenphoMsc04Adapter(),
   // Beurer SIG scales (BF720/BF105) advertise/expose Body Composition 0x181B
   // and would otherwise be grabbed by the Mi Scale 2 adapter, so match them
   // first (by BF720/BF105 name or Beurer company id 0x0611).
--- a/src/scales/renpho-msc04.ts
+++ b/src/scales/renpho-msc04.ts
@@ -1,0 +1,256 @@
+import type {
+  BleDeviceInfo,
+  ConnectionContext,
+  ScaleAdapterCore,
+  GattWiring,
+  ScaleReading,
+  UserProfile,
+  BodyComposition,
+} from '../interfaces/scale-adapter.js';
+import { uuid16, buildPayload } from './body-comp-helpers.js';
+import type { MatchDescriptor } from './match-descriptor.js';
+import { bleLog } from '../ble/types.js';
+
+// Service 0x1A10 — custom Renpho scale service (shared with ES-26BB)
+const CHR_CONTROL = uuid16(0x2a11); // write  — commands to scale
+const CHR_STATUS = uuid16(0x2a10); // notify — status / progress
+const CHR_RESULTS = uuid16(0x2a12); // indicate — final BIA measurement
+
+// 55-AA frame magic
+const MAGIC0 = 0x55;
+const MAGIC1 = 0xaa;
+
+// Request a measurement session.
+// Identical to ES-26BB start command: 55 AA 90 00 04 01 00 00 00 94
+const START_CMD = [0x55, 0xaa, 0x90, 0x00, 0x04, 0x01, 0x00, 0x00, 0x00, 0x94];
+
+// Ack a cached offline measurement so the scale stops replaying it on every
+// connect.  Last byte = sum(prev) & 0xFF = (55+AA+95+00+01+01) & FF = 96.
+const OFFLINE_ACK = [0x55, 0xaa, 0x95, 0x00, 0x01, 0x01, 0x96];
+
+// Command bytes on the indication characteristic
+const CMD_MEAS_SHORT = 0x25; // 36-byte payload (post-composition result)
+const CMD_MEAS_LONG = 0x26; // 40-byte payload (pre-composition + result)
+
+/**
+ * Extended scale reading that carries the pre-computed BIA metrics the
+ * R-MSC04 transmits in the indication frame.  Stored on the adapter instance
+ * between parseNotification() and computeMetrics().
+ */
+interface ExtendedBia {
+  bodyFatPct: number;
+  bmi: number;
+  skelMuscleMassKg: number;
+  boneMassKg: number;
+  visceralFat: number;
+}
+
+function isChecksumValid(data: Buffer): boolean {
+  if (data.length < 2) return false;
+  let sum = 0;
+  for (let i = 0; i < data.length - 1; i++) sum = (sum + data[i]) & 0xff;
+  return sum === data[data.length - 1];
+}
+
+/**
+ * Adapter for the Renpho R-MSC04 body composition scale.
+ *
+ * ─── GATT layout (service 0x1A10) ───────────────────────────────────────────
+ *   0x2A11  write    — send 55-AA commands (start, offline ack)
+ *   0x2A10  notify   — status / progress events (not parsed here)
+ *   0x2A12  indicate — final BIA measurement frame (55-AA cmd 0x25 / 0x26)
+ *
+ * ─── Session flow ───────────────────────────────────────────────────────────
+ *   1. Write START_CMD to 0x2A11 after connect.
+ *   2. Scale streams status notifications on 0x2A10 (ignored).
+ *   3. Scale sends final measurement indication on 0x2A12.
+ *   4. Ack with OFFLINE_ACK so cached frames are cleared from scale memory.
+ *
+ * ─── Measurement frame layout (55-AA frame on 0x2A12) ──────────────────────
+ *   [0-1]  magic: 55 AA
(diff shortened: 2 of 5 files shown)


KristianP26 (changes requested):
Thanks for this, it is genuinely useful groundwork and the central hypothesis is well motivated: the MorphoScan family appears to deliver the final measurement on the indicate characteristic `0x2A12`, while ES-CS20M only subscribes to notify `0x2A10`, which neatly explains #265's "subscribes, waits about 5s, disconnects, no weight" symptom (and lines up with the sibling #230 R-MSC02 note about a `0x2A12` indicate stream). Writing the start command once in `onConnected` instead of the ES-CS20M burst is also a sensible improvement.

Before this can merge I need a few changes, and I do not want it to auto-close #117 or #265 yet (reopen-if-regression pattern until a reporter confirms a real reading).

Critical:

1. The matcher may not actually fix #265. `matches()` only returns true on name `r-msc04` or manufacturer id `0x1a10`. But #265's device matched ES-CS20M by service UUID `1a10`, not by name. The new adapter deliberately (and correctly) does not claim service `1a10`, so the fix only works if BlueZ reports the GAP name as `R-MSC04` at match time. If the advertised name is empty (which is likely why it fell through to the service match), `matches()` returns false and the device still resolves to ES-CS20M. Could @vossitch post a `npm run scan` (or diagnose) dump showing the advertised name? And please add a test asserting `resolveAdapter({ localName: 'R-MSC04', serviceUuids: ['1a10'] })` picks R-MSC04 over ES-CS20M. The current collision fixture uses `serviceUuids: []`, which sidesteps the real collision.

2. The body-composition decode looks overfit. Weight and bone are read big-endian while impedance, fat, BMI and muscle are read little-endian within the same frame. Mixed endianness in one vendor frame is unusual and suggests the offsets and divisors were fit to hit target metrics. Impedance also decodes to about 181 Ohm (low for whole-body BIA), and the BMI field is internally inconsistent between the two captures. Only weight (BE u16 / 100) looks trustworthy. Please treat fat/muscle/bone/visceral/impedance as unverified pending a hardware run with a known body-fat reference from the Renpho app.

Should-fix:

3. `manufacturerId: 0x1a10` is semantically wrong: `0x1A10` is a GATT service UUID, not a Bluetooth SIG company identifier, and `matchesDescriptor` does not consult `manufacturerId` at runtime. Please follow the sibling convention: `match = { priority: 235, names: { exact: ['r-msc04'] } }` with `matches(d) { return matchesDescriptor(d, this.match); }`, and drop both `manufacturerId` and `custom: true`.

4. `tests/wizard/_test-config-ni.yaml` leaked into the PR. It is a transient file the wizard test writes and deletes; please remove it.

5. Please retarget the PR from `main` to `dev` (the flow is feature to dev, then release-please to main).

Minor: reset `this._bia = null` in `onConnected` (the registry holds one singleton, so a stale composition could pair with a new weight); the offline-ack on every 0x25/0x26 parse is unverified; skeletal muscle is fed into the total-muscle slot; and please update `docs/guide/supported-scales.md`.

Once (3), (4), (5) and a scan dump for (1) are in, I am happy to land it on `dev` as a provisional adapter, then close #117/#265 after a confirmed real reading. Thanks again for the solid start.

KristianP26:
@udochrist an update that directly affects this PR, and that should make it much easier to finish.

@joelr posted an HCI capture of a real R-MSC04 session on #117, and I decoded it. [Full writeup here.](https://github.com/KristianP26/ble-scale-sync/issues/117#issuecomment-4934834860) The relevant findings for your adapter:

1. **Framing is `55 AA | cmd | length (2B big endian) | payload | checksum`, where the checksum is the low byte of the sum of all preceding bytes.** I verified this on 13 frames. It also validates the existing ES-CS20M unlock command, so this is one protocol family.

2. **The final measurement arrives on a third, INDICATE characteristic**, not on the notify characteristic. The vendor app enables notifications on one CCCD and indications on another. This is why every R-MSC04 report so far shows the connection succeeding and then producing nothing. Any adapter for this scale has to subscribe to the indicate characteristic, which means declaring `characteristics` bindings rather than the single `charNotifyUuid` pair.

3. **Live weight** is cmd `0x21`, weight at payload bytes 3 and 4, big endian, divided by 100. **Final weight** is cmd `0x24` on the indicate characteristic, weight at payload bytes 4 and 5, same scaling. A capture showed a final of `0x2553` = 95.55 kg.

4. There is a handshake before the scale will stream: a ready frame, a **time sync** (big endian unix seconds, confirmed to the second against the capture wall clock), a user profile, a user name as raw ASCII, and a settings frame. Stored records are then dumped as fragmented indications, each one acknowledged with a `0xB6` frame before the next is sent.

This confirms my earlier review comment that the decode in this PR is fitted rather than derived. The mixed-endian weight parsing in particular does not survive contact with these frames.

My suggestion: rather than iterating on the current diff, wait for the UUID to handle map I asked @joelr and @vossitch for, then rebuild the adapter around the real framing. I am happy to take a fresh PR, or to write it myself and credit you, whichever you prefer. Either way the two blocking points from my earlier review still stand: `manufacturerId: 0x1a10` is a service UUID and not a Bluetooth SIG company identifier, and this should target `dev` rather than `main`.

Please also drop `tests/wizard/_test-config-ni.yaml`, which looks unrelated to this change.

KristianP26:
Status: I am not merging this one yet, and it needs a few changes first (my earlier review has the detail). In short:

1. It targets main; it should target dev, where all adapter work lands.
2. The match criteria may not reliably identify the R-MSC04 without confirming the advertised name, so it risks not fixing #265.
3. The frame decode looks overfit/unverified, the manufacturerId is a service UUID rather than a company id, and there are stray files (a wizard yaml and unrelated test edits) to drop.

The fastest path to landing this: retarget to dev, and post a scan dump plus one Android HCI snoop of a weigh-in so I can verify the match and the weight/impedance decode against real bytes. @joelr's working ESP bridge referenced in #117 is a good cross-check for the protocol. Once the decode is confirmed and the match is name-anchored, I am happy to merge.

KristianP26:
Thanks again @udochrist. I have shipped a weight-only R-MSC04 adapter to `dev` (commit f0dc658), and I want to be transparent about what I took from your PR and what I changed.

What your PR got right, and what I used: the GATT layout. Service 0x1A10 with write 0x2A11, notify 0x2A10, and indicate 0x2A12 for the final measurement. That matched @joelr's HCI capture in #117 exactly, and it is the part that actually fixes #265 (the ES-CS20M adapter never subscribed to the indicate characteristic). The exact-name `r-msc04` routing is also what I used.

What I changed, and why I did not merge this PR as-is:

1. The frame decode. Your cmd 0x25/0x26 body-composition field layout (impedance, BMI, bone mass, visceral fat at fixed offsets) could not be reconciled with the real capture, where the final weight arrives as a cmd 0x24 frame and the body-composition records are a separate cmd 0x26 dump gated behind a multi-write handshake. I did not want to ship a body-composition decode fitted to guessed offsets, so the new adapter is weight only, verified against joelr's 95.55 kg frame.
2. `manufacturerId: 0x1a10` is a service UUID, not a Bluetooth SIG company identifier, so it would never match manufacturer data. The name match alone is correct and sufficient.
3. It carried some unrelated files (a wizard test yaml).

I would still like your help finishing the body-composition side. If you can capture two or three sessions at known weights, I can decode the cmd 0x26 records and add fat, water, and muscle on top of what is now on `dev`. I am leaving this PR open and it is your call: if you want to carry the body-composition work forward, rebasing onto the new adapter would be the cleanest base; otherwise feel free to close it. The credit for confirming the GATT layout is yours either way.

udochrist:
Hi Kristian, sorry for the unresponsiveness. I‘m was/am on vacation and will follow up after my return.

KristianP26:
Status update, so this is not left hanging on you unfairly.

While this PR was in review, R-MSC04 support landed on `dev` through a separate weight-only adapter (`src/scales/renpho-msc04.ts`), and a reporter with the actual hardware has now confirmed it works end to end on `dev`: correct name match, the `0x2A11` write / `0x2A10` notify / `0x2A12` indicate layout, a clean 75.20 kg read, and an MQTT publish (see #265). So the core support this PR set out to add is already shipped and hardware-validated, which means this PR is superseded as it stands, and the change-requests I left earlier (the overfit decode, the `0x1A10` service-vs-company-id mix-up, the stray wizard/test yaml, and the conflicts against `dev`) no longer need chasing.

I did not want to close your PR out from under you. Two ways forward, your call:

1. If you are happy that the feature is covered, feel free to close this. No hard feelings, and thank you for the push on #117; it genuinely helped move it along.
2. If you want to keep contributing, the one piece still open is body composition. The shipped adapter is weight-only because the final impedance / body-comp arrives on the `0x2A12` indicate and I have not decoded that frame yet. A PR that decodes `0x2A12` (ideally backed by an HCI capture of a weigh-in, so the transform is verifiable rather than inferred) would be very welcome and would land.

Either way, appreciate the work.

KristianP26:
Closing this one, with thanks rather than as a rejection.

The R-MSC04 support this PR set out to add shipped on `dev` through `src/scales/renpho-msc04.ts` and was hardware-confirmed by a reporter in #265, and the branch has since gone conflicting against `dev`, so there is nothing left here that can be merged as it stands. Your investigation is what pointed at the right service layout in the first place, and that is in the shipped adapter.

If you still want to contribute the piece that is genuinely missing, it is the body-composition decode: the shipped adapter is weight-only because the final frame arrives on the indicate characteristic `0x2A12` and I have no capture that maps its bytes to fat and impedance. A small PR that decodes that frame, or just the raw hex of one weigh-in from your scale, would finish the job. #117 and #265 stay open for exactly that.

No rush and no obligation, and reopening this PR is fine if you would rather rebase it.

PR #267, opened 2026-07-05
fix(renpho): perform consent handshake so the ES-WBE28 streams readings

junaidk:
## Problem

The **Renpho ES-WBE28** ("Renpho-Scale", Elis 1 family) connects but never produces a reading — the log shows `Matched adapter: Renpho ES-WBE28` → `Subscribed to notifications. Step on the scale.` → (after the step-on timeout) `Scale disconnected before reading completed`.

Root cause: this scale speaks **standard Bluetooth SIG GATT** (Weight Scale `0x2A9D`, Body Composition `0x2A9C`, User Data `0x2A9F`) and stays completely **silent until a registered user consents** through the User Data Service. The previous adapter only:

- re-sent a single vendor `MAGIC0` write (`10 01 00 11`) to `0xFFE2` every 3s via the legacy `Unlockable` path, and
- subscribed to `0x2A9D`.

It never sent the rest of the vendor handshake, never sent UCP consent, and never wrote the user profile — so the scale never unlocked and eventually dropped the link.

A second latent bug: `parseNotification()` required `data[0] === 0x2E` as a frame marker, but real frames from this device begin with a standard SIG **flags** byte (e.g. `0x4E`), so every measurement would have been dropped even if the scale did stream.

## Fix

Rewrite the adapter to replay the app's init through the `onConnected` hook, modeled on the existing Beurer BF720 SIG-consent adapter:

1. Auto-subscribe the vendor (`0xFFE1`) + User Control Point (`0x2A9F`) channels.
2. Vendor handshake — three writes to `0xFFE2` (`10 01 00 11`, `03 00 01 04`, `19 00 19`).
3. UCP consent — `02 AA 0F 27` (user index 170, code 9999).
4. Profile writes (gender / height / DOB / age + vendor `0x2AFF`), each guarded by `availableChars` so a firmware variant missing one doesn't abort init.
5. Enable the weight + body-composition notifications **last** (order matters — the app enables these CCCDs only after consent).

Parsing now decodes standard SIG Weight Measurement + Body Composition frames (with RENPHO's documented `0.05 kg` resolution deviation) and extracts **impedance**. The link is held open briefly (`completionHoldMs`) so the multi-packet composition indications carrying impedance can land, and body fat is recomputed host-side from impedance (BIA) rather than trusting the scale's per-profile derived values.

The consent user index (170) and code (9999) are the returning-user values captured from the RENPHO app and are fixed for this device family.

## Testing

- Rewrote `tests/scales/renpho.test.ts` (17 tests): `matches()`, the full `onConnected` handshake ordering, consent-char-missing error, firmware-variant profile skipping, SIG weight/impedance decoding (using real captured frames → **94.9 kg** / **332.8 Ω**), imperial→kg, UCP-response handling, truncated-frame safety, and BIA `computeMetrics`.
- `npm test` (full scales suite: 617 passing), `npx tsc --noEmit`, `npm run lint`, `npm run format:check` all clean.

Verified against the physical device's protocol capture; the byte-level test fixtures come straight from that capture. I don't have a second ES-WBE28 to check consent-code portability, but the values match the app's registration for this model family.

### DIFF
--- a/src/scales/renpho.ts
+++ b/src/scales/renpho.ts
@@ -1,43 +1,98 @@
 import type {
   BleDeviceInfo,
+  CharacteristicBinding,
+  ConnectionContext,
   ScaleAdapterCore,
   GattWiring,
-  Unlockable,
+  MultiCharNotify,
+  HoldForComposition,
   ScaleReading,
   UserProfile,
   BodyComposition,
 } from '../interfaces/scale-adapter.js';
-import { uuid16, buildPayload } from './body-comp-helpers.js';
+import { uuid16, buildPayload, computeBiaFat, type ScaleBodyComp } from './body-comp-helpers.js';
+import { bleLog } from '../ble/types.js';
 import type { MatchDescriptor } from './match-descriptor.js';
 
 /**
- * Ported from openScale's RenphoHandler.kt
+ * Handler for RENPHO ES-WBE28 / "Renpho-Scale" (Elis 1 family).
  *
- * Handler for RENPHO ES-WBE28 — a newer Renpho model that uses standard BLE
- * services (Body Composition 0x181B, User Data 0x181C, Weight Scale 0x181D,
- * Current Time 0x1805) but with vendor-specific payload encoding.
+ * Reverse-engineered from an Android btsnoop HCI capture of the physical device
+ * and verified against it. Despite openScale's older RenphoHandler.kt treating
+ * this as a proprietary weight-only device, the ES-WBE28 speaks pure Bluetooth
+ * SIG GATT:
  *
- * Weight is published on the Weight Measurement characteristic (0x2A9D) in a
- * proprietary format: data[0] must equal 0x2E, weight = LE uint16 at [1-2] / 20.0.
+ *   Weight Scale       0x181D → Weight Measurement       0x2A9D (notify)
+ *   Body Composition   0x181B → Body Comp. Measurement    0x2A9C (indicate)
+ *   User Data          0x181C → User Control Point        0x2A9F (consent)
  *
- * Vendor handshake:
- *   MAGIC0 [0x10, 0x01, 0x00, 0x11] → written to 0xFFE2
- *   MAGIC1 [0x03, 0x00, 0x01, 0x04] → written to 0xFFE2
- *   MAGIC_UCP [0x02, 0xAA, 0x0F, 0x27] → written to 0x2A9F (consent user 0xAA, code 9999)
+ * The scale stays completely SILENT until a registered user "consents" through
+ * the User Data Service. Merely subscribing to 0x2A9D and poking the vendor
+ * channel (the previous implementation) never unlocks it, so the scale connects
+ * and then disconnects without ever streaming. We must replay the app's init:
  *
- * Mutual exclusion with QnScaleAdapter:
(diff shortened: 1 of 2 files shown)


KristianP26:
This is an excellent writeup and the implementation tracks the Beurer BF720 SIG-consent adapter closely. The handshake to consent to profile to subscribe-last ordering, the multi-char dispatch, and the hold-for-impedance logic all wire up correctly against `waitForRawReading`. I hand-verified the body-composition field walk against your `BCS_PACKET2` fixture (impedance lands at offset 7, so 332.8 Ohm, which is the SIG 0.1 Ohm resolution) and the 0.05 kg factor is consistent with the old `/20.0` Renpho scaling, so the decode reasoning is solid. Since the previous adapter required `data[0] === 0x2E` (real SIG frames start with a flags byte), it never actually parsed a frame, so regression risk to existing users is effectively zero.

One thing I want to confirm before merging to `main`: the ES-WBE28 uses the SIG User Data Service (`0x2A9F` plus the profile chars), which on the Beurer sibling required `requiresBonding = true` because BlueZ will not enable those CCCDs on an unencrypted link. Your capture is from Android, which auto-bonds and can hide that requirement. On Linux/node-ble we only bond when the adapter opts in. Could you run it once on your target platform with `DEBUG=true` and confirm you get past the `0x2A9F` subscribe and see the consent accepted plus a real weight? If it fails with a "requires a bonded/encrypted link" error, we will just add `requiresBonding = true`.

Two smaller notes, neither blocking:
- The imperial path (`LB_PER_UNIT = 0.1`) is only tested against a synthetic frame, so it is unverified against a real imperial capture.
- A few code comments use em dashes, which the repo style avoids; I may sweep those on merge, no action needed from you.

The code is merge-ready to `dev`. I would like the one DEBUG-log run on real hardware before it ships to `main`. Thanks again, this is a high-quality contribution.

KristianP26:
Merged to dev. Thanks, this is a solid rewrite and closely mirrors the Beurer BF720 SIG-consent adapter. I reviewed the regression surface carefully: the matcher is unchanged so no currently-working Renpho model is affected, QN still wins the shared cases by priority, the weight scaling is identical to before, and a failed handshake now rejects and retries cleanly rather than crashing.

I applied one follow-up on top of your branch (commit f4d7d5c): onConnected only hard-requires the UCP characteristic now, and the vendor 0xFFE2 writes are guarded by availableChars. Your matcher rejects devices advertising the 0xFFE0 service, so a firmware variant may not expose 0xFFE2, and consent alone unlocks the stream, so requiring 0xFFE2 was self-contradictory.

Three things to validate on real ES-WBE28 hardware, since I could not confirm them without the device:

1. Bonding: 0x2A9F lives in the SIG User Data Service, and the BF720 sets requiresBonding for exactly that reason (its CCCDs need an encrypted link). Your Android capture likely bonded implicitly. If subscribing to 0x2A9F or the UCP write fails on BlueZ, we may need requiresBonding = true here too. Please test over the native node-ble handler and report whether it streams.
2. Profile writes: confirm the scale actually needs the gender/height/DOB/age writes to unlock, or whether consent alone is enough (we recompute body comp from impedance regardless).
3. The imperial (lb) weight factor is extrapolated, not from a captured lb frame. A known-weight run with the scale in lb would confirm it.

A known-weight run (kg) that produces a correct reading is the main thing. Ships in the next release.

PR #268, opened 2026-07-05
feat(scales): add Hutbit smart scale adapter

### LINKED ISSUE #254: feat(scales): add Hutbit Smart Scale (218008/WL292) — decoded Lefu/Fitdays FFB0 "AC02" protocol
### Description

Support request for the **Hutbit Smart Scale** (model **218008**, B/N **WL292**, app: **Fitdays**, OEM: Lefu).

I decoded the full protocol from two Android Fitdays HCI snoops with known weights. It's a Lefu FFB0 scale, but a **different, simpler protocol than the Robi S9** (#228): fixed 8-byte frames, weight streamed on FFB2, and **no app-identity token** in the handshake.

**Device**
- Advertised name: `Hutbit Scale`
- MAC / identity: `03:B3:EC:93:B8:7E`
- Manufacturer ID: `0x02AC` (684); mfr data = MAC reversed + a status byte
- Services: `ffb0` (+ an unused `d618` service)

**GATT (service 0xFFB0)**
- `FFB1` (handle 0x001c) — Write Without Response — commands phone→scale
- `FFB2` (handle 0x0018) — Notify — weight + result stream scale→phone
- `FFB3` (0x0020) — Indicate — **unused** in the entire capture

**Frame format (fixed 8 bytes)**
```
AC 02 | D0 D1 D2 D3 | STATUS | CKSUM
CKSUM = (D0 + D1 + D2 + D3 + STATUS) & 0xFF      // verified on every frame
```

**Handshake** (phone → FFB1, write-no-response), after enabling FFB2 notify (`0100` to CCCD 0x0019):
```
ac02 fa01 0000 cc c7
ac02 fb02 1fa5 cc 8d     // 1fa5 = fixed config value, also echoed in the result frame
ac02 fde2 0101 cc ad
ac02 fc01 0000 cc c9
ac02 fe06 0000 cc d0     // poll/keepalive, repeated by the app through the session
```

**Weight** (FFB2 notify): `AC 02 [weight_u16_BE] 00 00 [STATUS] [CKSUM]`, **weight_kg = u16 / 10**.
STATUS `0xCE` = measuring/unstable, **`0xCA` = stable/final**.

**Ground-truth captures** (weight settling, then stable `0xCA` frames):
- Session 1 (Fitdays showed **84.1 kg**): `ac0203490000ca16` → `0x0349` = 841 → 84.1 kg ✓
- Session 2 (Fitdays showed **84.x kg**): settles to `0347`/`0348`/`034a` (83.9–84.2) ✓

**Body composition:** a 40-byte type-`FF` result frame (split over two notifies, e.g. `ac02ff000221e2…0349…` + `…01021fa5…`) also carries weight + impedance-ish fields, **but this unit's bioimpedance sensor is unreliable** (it reported 0.4–0.7 % body fat across many tries). I'd suggest treating it as **weight-only and deriving body composition via the existing BIA pipeline**, same as the Renpho / Robi S9 adapters.

### Use Case

Two-user household syncing weight to Home Assistant via MQTT over an existing ESPHome Bluetooth proxy (no dedicated ESP32). Over the esphome-proxy transport the add-on only GATT-connects when an adapter claims the advertisement, so an unrecognised Hutbit advert currently produces nothing — a dedicated adapter is needed.

### Alternatives Considered

- `homeassistant_chipsea_scale` — matches `fff0` / mfr 4298, doesn't fit Hutbit's `ffb0` / `0x02AC`.
- Existing MGB / Robi S9 FFB0 adapters — don't recognise the advert (Robi S9's nameless match needs the FFB3 char, which Hutbit doesn't use), so nothing connects.

I've written a **draft adapter** (matches `hutbit` / `ffb0`+`0x02AC`, replays the FFB1 handshake, parses FFB2 weight, validates the checksum, weight-only → BIA). Happy to open a PR against `dev`, and/or attach the raw `btsnoop_hci.log` captures (two weigh-ins at known weights) if useful — just say which you prefer.

bondesen:
Closes #254.

Adds an adapter for the **Hutbit Smart Scale** (model **218008** / B/N **WL292**, Fitdays app, Lefu OEM), decoded from two Android Fitdays HCI snoops taken at known weights.

### Protocol (Lefu FFB0 "AC02")
Shares service `0xFFB0` with the Robi S9 / openScale MGB families but speaks a **simpler, fixed 8-byte protocol** — not the Robi's 20-byte one:

```
AC 02 | D0 D1 D2 D3 | STATUS | CKSUM
CKSUM = (D0 + D1 + D2 + D3 + STATUS) & 0xFF   // verified on every frame
```

- **Handshake** replayed on `FFB1` (write-no-response) after enabling `FFB2` notifications. Frames are fixed and carry **no app-identity token**, so they're safe to replay verbatim.
- **Weight** streamed on `FFB2` notify: `AC 02 [weight_u16_BE] 00 00 [STATUS] [CKSUM]`, `weight_kg = u16 / 10`. `STATUS 0xCE` = measuring/unstable, **`0xCA` = stable/final**.
- Ground truth: `ac0203490000ca16` → `0x0349` = 841 → **84.1 kg** ✓ (matched the Fitdays reading).
- `FFB3` is present but **unused**; the unit's bioimpedance is unreliable (0.4–0.7 % body fat), so this is **weight-only** → body composition is derived via the shared BIA/BMI pipeline, same as the Robi S9 / Renpho adapters.

### Changes
- `src/scales/hutbit.ts` — `HutbitAdapter` (`ScaleAdapterCore` + `GattWiring`), matches by the `Hutbit` name, `custom: true`, unique `priority: 35`, checksum-validated `FFB2` parse, weight-only → `impedance: 0`.
- `src/scales/index.ts` — registered before the generic MGB `FFB0` fallback.
- `tests/scales/hutbit.test.ts` — matches/registry-resolution, handshake order + checksum, weight decode from the real capture, measuring-frame + bad-checksum rejection, metrics range check.
- `tests/scales/registry-collision.test.ts` — added the required first-match fixture.

Matching is intentionally **name-only** — the nameless `FFB0` space is left to Robi S9 (FFB3 result char) and the MGB fallback to avoid shadowing them.

### Verification (local)
`npm test` (1830 tests), `npm run lint`, `npm run format:check`, `npx tsc --noEmit` — all green.

I still have the two raw `btsnoop_hci.log` captures (weigh-ins at known weights) and can attach them or adjust anything to your conventions — just say the word.

### DIFF
--- a/src/scales/hutbit.ts
+++ b/src/scales/hutbit.ts
@@ -1,0 +1,127 @@
+import type {
+  BleDeviceInfo,
+  CharacteristicBinding,
+  ConnectionContext,
+  ScaleAdapterCore,
+  GattWiring,
+  ScaleReading,
+  UserProfile,
+  BodyComposition,
+} from '../interfaces/scale-adapter.js';
+import { uuid16, buildPayload } from './body-comp-helpers.js';
+import { bleLog } from '../ble/types.js';
+import type { MatchDescriptor } from './match-descriptor.js';
+
+// ─── Hutbit Smart Scale (Lefu / Fitdays FFB0 "AC02" 8-byte protocol) ─────────
+
+const CHR_FFB1 = uuid16(0xffb1); // write (handshake, phone→scale)
+const CHR_FFB2 = uuid16(0xffb2); // notify (weight stream, scale→phone)
+
+/**
+ * Handshake replayed on FFB1 (write-without-response) after enabling FFB2
+ * notifications. Fixed 8-byte AC02 frames decoded from two Fitdays HCI snoops
+ * (#254). Unlike the Robi S9's 20-byte protocol, every frame is a fixed
+ * `AC 02 | D0 D1 D2 D3 | STATUS | CKSUM` with a plain additive checksum and NO
+ * app-identity token, so the frames are stable and safe to replay verbatim.
+ * `ac02fe060000ccd0` is the poll/keepalive the app repeats through the session.
+ */
+const HANDSHAKE: string[] = [
+  'ac02fa010000ccc7',
+  'ac02fb021fa5cc8d',
+  'ac02fde20101ccad',
+  'ac02fc010000ccc9',
+  'ac02fe060000ccd0',
+];
+
+const FRAME_LEN = 8;
+const FRAME_HEADER0 = 0xac;
+const FRAME_HEADER1 = 0x02;
+const STATUS_STABLE = 0xca; // final/stable reading (0xCE = measuring/unstable)
+const WEIGHT_DIV = 10; // weight = u16 BE / 10 → kg
+
+/** Additive checksum over D0..STATUS (bytes 2..6), matching the vendor frames. */
+function frameChecksum(data: Buffer): number {
+  return (data[2] + data[3] + data[4] + data[5] + data[6]) & 0xff;
+}
+
+/**
+ * Adapter for the Hutbit Smart Scale (model 218008 / WL292, Fitdays app, Lefu
+ * OEM). It shares service 0xFFB0 with the Robi S9 and openScale MGB families
+ * but speaks a simpler, fixed 8-byte "AC02" protocol: after enabling FFB2
+ * notifications the phone replays a short FFB1 handshake, then the scale streams
+ * `AC 02 [weight_u16_BE] 00 00 [STATUS] [CKSUM]` frames on FFB2, where
+ * STATUS 0xCE = measuring/unstable and 0xCA = stable/final. Weight = u16 / 10 kg.
+ *
+ * Decoded from two known-weight HCI snoops (#254): `ac0203490000ca16`
+ * = 0x0349 = 841 → 84.1 kg. The unit's bioimpedance is unreliable (it reported
+ * 0.4–0.7 % body fat), so it is treated as weight-only and body composition is
+ * derived via the shared BIA/BMI pipeline, same as the Robi S9 / Renpho adapters.
+ */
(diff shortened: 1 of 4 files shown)


KristianP26:
Thanks, this is an excellent PR. The protocol write-up in #254 made review straightforward and the adapter matches it exactly: checksum over `D0..STATUS`, weight `u16BE / 10`, `0xCA`-gated final frame, and the verbatim FFB1 handshake. I verified the ground-truth frame `ac0203490000ca16` decodes to 84.1 kg with a valid checksum, and the measuring/`0xCE` frame is correctly dropped.

Matcher precedence is right too: `priority: 35` sits cleanly between MGB (30) and Robi S9 (40), `custom: true` keeps it out of the overlap check, and name-only matching leaves the nameless FFB0 space to Robi (which needs FFB3) and the MGB fallback. Good test coverage from the real captures.

Two small things before I merge:

1. The handshake writes to FFB1 pass `withResponse: true`, but you documented FFB1 as Write Without Response, and MGB uses `false` on the same characteristic. Since you are running over an ESPHome proxy (which forwards that flag to the underlying write), please switch the handshake writes to `false`. It matches the characteristic property and de-risks the proxy path.
2. Minor: there is an en dash in the `0.4-0.7 % body fat` doc comment. This repo avoids en and em dashes, so please reword (for example `0.4 to 0.7 %`).

Once those are in I will merge (it closes #254). The failing `build-and-push` check is an unrelated transient registry 503, not your change. And yes please attach the two `btsnoop_hci.log` captures to #254 for the archive. Thanks again.

PR #274, opened 2026-07-11
fix(firmware): enable BLE CCCD notifications in start_notify

marcelorodrigo:
Enables BLE notifications on the peripheral by writing the Client Characteristic Configuration Descriptor.

The latest retest logs in #231 showed:

```
Subscribe: notify enabled for 0000fff1...
Subscribe: notify enabled for 0000ae02...
Notify loop error (0000fff1...):
Notify loop error (0000ae02...):
```

The empty error message after the colon is `asyncio.TimeoutError()` from `char.notified()`. aioble's `notified()` only consumes events; it does not turn on the CCCD bit. Without that write, the scale never starts sending notifications, so the loop waits 10s and times out with no data.

This change adds `char.subscribe(notify=/indicate=)` in `BleBridge.start_notify()` based on the characteristic properties, and adds REPL debug logs so the CCCD step is visible in future logs. A regression test covers notify, indicate, and missing-characteristic paths.

All 93 firmware tests pass. The unrelated `tests/config/watch.test.ts` failure is pre-existing on `dev`.

Relates to #231

### DIFF
--- a/firmware/ble_bridge.py
+++ b/firmware/ble_bridge.py
@@ -572,10 +572,28 @@ class BleBridge:
         return {"chars": chars_info}
 
     async def start_notify(self, uuid_str, publish_fn):
-        """Start forwarding notifications from a characteristic via publish_fn."""
+        """Start forwarding notifications from a characteristic via publish_fn.
+
+        This also writes the CCCD (Client Characteristic Configuration
+        Descriptor) so the peripheral actually sends notifications. Without this
+        step aioble's char.notified() just waits forever on an empty queue and
+        times out with no data (#231)."""
         char = self._chars.get(uuid_str)
         if not char:
             return
+
+        # Enable notify/indicate on the peripheral. aioble's notified() only
+        # consumes events; it does not turn on the CCCD bit itself.
+        print(f"Subscribe: enabling CCCD for {uuid_str}")
+        try:
+            await char.subscribe(
+                notify=bool(char.properties & bluetooth.FLAG_NOTIFY),
+                indicate=bool(char.properties & bluetooth.FLAG_INDICATE),
+            )
+        except Exception as e:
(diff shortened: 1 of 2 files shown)


KristianP26 (approved):
Thanks Marcelo, this is a correct and well diagnosed fix. You have the root cause exactly right: aioble's `char.notified()` only drains the event queue, it never writes the CCCD, so without `char.subscribe()` the peripheral is never told to start sending and the loop just waits out its 10s timeout with the empty `asyncio.TimeoutError` you saw in the #231 logs. Deriving `notify=`/`indicate=` from `char.properties` is the right way to pick the descriptor bit, and it lines up with how the discovery path at ble_bridge.py:552/556 already reads those flags. The regression test covering notify, indicate, and the missing-char no-op is a nice touch.

One non-blocking point on error handling. `start_notify` now `raise`s if `subscribe()` fails, and in the eager path (`main.py:237-242`) that call runs in a bare loop before `client.publish(topic("connected"), ...)`. So a single characteristic that rejects its CCCD write would abort the whole autonomous-connect flow and the host would never receive the connected payload, where today a bad char is a silent skip. For the QN target this should not trigger, and failing fast on a genuine CCCD error is defensible, but consider a per-characteristic try/except that logs and continues so one uncooperative char cannot drop the entire session. Your call, happy to merge either way.

Scope note for #231: this fixes the notify stage, but your latest REPL log on the issue shows the flow stalling at the connect stage itself ("detected after 2000ms, connect" then silence, no notify errors yet), which looks like a separate problem from the missing CCCD. Once this is merged to dev, could you retest end to end and confirm whether weight frames now actually stream, or whether it still stalls before notifications ever start? That will tell us if #231 needs a second fix on the connect path.

Approving. I will merge to dev.

PR #279, opened 2026-07-16
fix(scales): claim OEM/rebranded Hutbit ("SWAN") via the 0x02AC manufacturer signature

### LINKED ISSUE #278: Hutbit 218008 (#254/#268) advertises as "SWAN" — Hutbit adapter never claims it, Robi S9 wins re-resolution and the scale drops the link
Follow-up to #254 / #268 (I'm the Hutbit reporter/contributor). With the adapter released in **v1.20.0** the scale still never reads — and live testing today shows why. Two stacked matching problems:

## 1. The scale advertises as "SWAN", not "Hutbit Scale"

The same physical unit (model 218008/WL292, identity MAC `03:B3:EC:93:B8:7E`) now advertises with `Complete Local Name: SWAN` — presumably Lefu OEM stock branding. nRF Connect capture today:

```
Connectable: Yes
Advertising type: Legacy
Adv. interval: 461 ms (while active) / 1005 ms (idle)
Flags: LE General Discoverable, BR/EDR Not Supported
Complete list of 16-bit Service UUIDs: 0xD618, 0xFFB0
Manufacturer data: Company: RTB Elektronik GmbH &amp; Co. KG <0x02AC>
  0x7EB893ECB30301   (active — last byte 01)
  0x7EB893ECB30300   (idle   — last byte 00)
Complete Local Name: SWAN
Raw: 0x020106 0503 18D6 B0FF 0AFF AC02 7EB893ECB303 01 0509 5357414E
```

The manufacturer payload is the device's own MAC **reversed** + a status byte — the same 0x02AC signature documented in #254. So `HutbitAdapter.matches()` (name-includes-'hutbit' only) never claims it.

## 2. What actually happens on v1.20.0 (HA add-on, esphome-proxy): MGB claims → Robi S9 steals re-resolution → scale drops the link, forever

```
07:56:18.621 [BLE] Matched: MGB (Swan/Icomon/YG) (03:B3:EC:93:B8:7E); opening GATT via ESPHome proxy
07:56:19.585 [BLE] Re-resolved adapter after GATT discovery: MGB (Swan/Icomon/YG) -> Robi S9 (03:B3:EC:93:B8:7E)
07:56:19.587 [BLE:debug] Subscribing to 0000ffb2... / 0000ffb3...
07:56:19.603 [BLE] Subscribed to 2 notification(s). Step on the scale.
07:56:19.667 [BLE:debug] Calling adapter.onConnected()
   ... link drops, and the cycle repeats every ~6 s while the scale is awake
```

A full weigh-in (display locked at 82.4 kg) produced zero frames — the Hutbit rejects the Robi 20-byte handshake and disconnects, exactly the failure mode #228 described for MGB-init on a Lefu scale.

Two contributing observations:

- **The local name never survives the ESPHome proxy.** Robi's `matches()` has a swan-name guard, but at re-resolution `device.localName` is empty (the proxy advertisement's `name` field was empty throughout — likely because the name is in the scan response), so Robi's *nameless* ffb0+ffb3 path claims it. The Hutbit exposes FFB3 (unused), so this always happens.
- Name-based matching alone is therefore fragile for this scale on the esphome-proxy transport, regardless of branding.

## Proposed fix

Match the Hutbit on its **manufacturer-data signature** instead of relying on the name: company id `0x02AC` + 7-byte payload ending in a `0x00`/`0x01` status byte (payload = reversed MAC), alongside the advertised `ffb0` service. That signal survives the proxy transport and the OEM rebranding. Additionally, guard Robi S9's nameless ffb0+ffb3 path against devices carrying that signature so re-resolution can't steal the connection (Hutbit prio 35 < Robi 40).

PR incoming from my fork with exactly that + tests. Happy to run live captures/tests on this unit — it reproduces within seconds.

bondesen:
Fixes #278 (follow-up to #254 / #268 — same unit, live-tested today on v1.20.0).

## Problem

The Hutbit 218008 ships with Lefu OEM stock branding: it advertises **`SWAN`** instead of `Hutbit Scale` — and over the ESPHome proxy the local name arrives **empty** entirely (it lives in the scan response, which the proxy advertisement doesn't carry). The name-only `matches()` therefore never claims the unit. Observed chain on v1.20.0:

```
Matched: MGB (Swan/Icomon/YG) (03:B3:EC:93:B8:7E); opening GATT via ESPHome proxy
Re-resolved adapter after GATT discovery: MGB (Swan/Icomon/YG) -> Robi S9
Subscribed to 2 notification(s). Step on the scale.
   ... scale rejects the Robi handshake and drops the link; repeats every ~6 s
```

Robi's swan-name guard can't fire at re-resolution because `localName` is empty there, so its nameless ffb0+ffb3 path claims the Hutbit's (unused) FFB3. A full weigh-in produced zero frames.

## Fix

Match on the stable fingerprint from the advertisement instead of the name: **manufacturer data company id `0x02AC` whose 7-byte payload is the device's own MAC reversed plus a status byte (`0x00` idle / `0x01` active)** — the signature documented in #254 — combined with the advertised `ffb0` service. nRF capture of both variants today:

```
Company: <0x02AC>  0x7EB893ECB30301   (active)
Company: <0x02AC>  0x7EB893ECB30300   (idle)
Services: 0xD618, 0xFFB0 · Legacy, connectable · name: SWAN
```

- `hutbit.ts`: `matches()` claims name-'hutbit' (unchanged) **or** ffb0 + the 0x02AC signature; exports `hasHutbitSignature()`; descriptor gains `manufacturerId`.
- `robi-s9.ts`: the nameless ffb0+ffb3 claim now excludes devices carrying the Hutbit signature, so post-discovery re-resolution stays on Hutbit (prio 35 < Robi's 40 makes the guard necessary). Named/real Robi units are unaffected, and nameless FFB3 units **without** the signature still resolve to Robi (covered by a test).
- The broader nameless FFB0 space is deliberately still left to Robi/MGB — without the 0x02AC signature nothing changes.

## Tests

7 new cases in `tests/scales/hutbit.test.ts` from the real capture: SWAN-named + empty-name (proxy) claims, idle/active payload variants, malformed-payload rejection, registry resolution (broadcast → Hutbit, not MGB), post-discovery re-resolution (→ Hutbit, not Robi), and the Robi non-shadowing control.

`vitest run`: 1907/1908 passing — the one failure is `tests/config/watch.test.ts > ignores edits to other files in the same directory`, a macOS `fs.watch` sibling-event flake unrelated to this change.

I'll live-test on the physical unit (HAOS add-on, esphome-proxy) as soon as a build is available — it reproduces within seconds here.

### DIFF
--- a/src/scales/hutbit.ts
+++ b/src/scales/hutbit.ts
@@ -17,6 +17,26 @@ import type { MatchDescriptor } from './match-descriptor.js';
 const CHR_FFB1 = uuid16(0xffb1); // write (handshake, phone→scale)
 const CHR_FFB2 = uuid16(0xffb2); // notify (weight stream, scale→phone)
 
+/** Lefu/Fitdays company id carried in the Hutbit's manufacturer data. */
+const MANUFACTURER_ID = 0x02ac;
+
+/**
+ * True when the advertisement carries the Hutbit's manufacturer-data signature:
+ * company id 0x02AC with a 7-byte payload of the device's own MAC reversed plus
+ * a status byte (0x00 idle / 0x01 active), as documented in #254. Exported so
+ * the Robi S9's nameless FFB0+FFB3 claim can exclude Hutbit units (#278): the
+ * Hutbit exposes an (unused) FFB3, and the local name does not survive every
+ * transport (the ESPHome proxy delivers advertisements with an empty name), so
+ * without this signature check the Robi adapter wins post-discovery
+ * re-resolution and replays a handshake the Hutbit rejects.
+ */
+export function hasHutbitSignature(device: BleDeviceInfo): boolean {
+  const m = device.manufacturerData;
+  return (
+    m?.id === MANUFACTURER_ID && m.data.length === 7 && (m.data[6] === 0x00 || m.data[6] === 0x01)
+  );
+}
+
 /**
  * Handshake replayed on FFB1 (write-without-response) after enabling FFB2
  * notifications. Fixed 8-byte AC02 frames decoded from two Fitdays HCI snoops
@@ -64,6 +84,7 @@ export class HutbitAdapter implements ScaleAdapterCore, GattWiring {
     custom: true,
     names: { includes: ['hutbit'] },
     serviceUuids: ['ffb0'],
+    manufacturerId: MANUFACTURER_ID,
   };
   readonly charNotifyUuid = CHR_FFB2;
   readonly charWriteUuid = CHR_FFB1;
@@ -77,12 +98,20 @@ export class HutbitAdapter implements ScaleAdapterCore, GattWiring {
   private final = false;
 
   matches(device: BleDeviceInfo): boolean {
-    // Advertised name is "Hutbit Scale"; match by name only. The nameless FFB0
-    // space is deliberately left to the Robi S9 (FFB3 result char) and the MGB
-    // fallback — the Hutbit exposes no unique post-discovery characteristic
-    // (FFB3 is present but unused), so claiming nameless FFB0 here would risk
-    // shadowing those adapters.
-    return (device.localName || '').toLowerCase().includes('hutbit');
+    // Branded units advertise "Hutbit Scale". Lefu OEM stock units advertise a
+    // generic name instead (observed: "SWAN", #278) — and over the ESPHome
+    // proxy the local name arrives empty entirely — so the name alone is not
+    // enough.
+    if ((device.localName || '').toLowerCase().includes('hutbit')) return true;
+
+    // OEM/rebranded units: claim on the manufacturer-data signature (0x02AC +
+    // reversed-MAC payload + status byte) combined with the advertised FFB0
+    // vendor service. This deliberately does NOT claim the broader nameless
+    // FFB0 space — without the 0x02AC signature that space still belongs to
+    // the Robi S9 (FFB3 result char) and the MGB fallback.
+    const uuids = (device.serviceUuids || []).map((u) => u.toLowerCase());
+    const hasFfb0 = uuids.some((u) => u === 'ffb0' || u === uuid16(0xffb0));
+    return hasFfb0 && hasHutbitSignature(device);
   }
 
   async onConnected(ctx: ConnectionContext): Promise<void> {
--- a/src/scales/robi-s9.ts
+++ b/src/scales/robi-s9.ts
@@ -11,6 +11,7 @@ import type {
 } from '../interfaces/scale-adapter.js';
 import { uuid16, buildPayload } from './body-comp-helpers.js';
 import { bleLog } from '../ble/types.js';
+import { hasHutbitSignature } from './hutbit.js';
 import type { MatchDescriptor } from './match-descriptor.js';
 
 // ─── Robi S9 (Lefu / Fitdays FFB0-new protocol) ─────────────────────────────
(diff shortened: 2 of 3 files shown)


KristianP26 (approved):
Thanks for this, and especially for going back to the physical unit and capturing the raw advertisement. The diagnosis is correct and I re-derived every step of it against the code before writing this:

- `hutbit.ts` `matches()` was name-only, so an OEM unit advertising `SWAN` was never claimed.
- `mgb.ts:31` claims the bare `ffb0` service through the descriptor (OR semantics in `match-descriptor.ts:82`), which is why MGB won the pre-connect match with an empty name.
- The esphome watcher re-resolve then handed the device to `robi-s9.ts:100-107`, whose nameless FFB0+FFB3 path cannot see your swan-name guard because `localName` is empty on that transport.

Your reading that the name lives in the scan response is right, and better supported than the alternatives. Your raw dump decomposes as flags 3B, complete 16-bit service list 6B (d618, ffb0), manufacturer data 11B, complete local name 6B. nRF Connect concatenates ADV and SCAN_RSP in that field, so the natural split is a 20-byte ADV plus a 6-byte SCAN_RSP carrying the name. That explains the empty name over a passive-scanning proxy and it puts the 0x02AC data and d618 in the ADV, where `matches()` can see them. Your own log corroborates it: MGB matched pre-connect with an empty name, and `matchesDescriptor` skips every name predicate when the name is empty, so it matched on `ffb0`, which proves the service list reaches `matches()` on that transport.

Keeping Hutbit at priority 35 and putting the guard after the `robi` name check is the right call, and I want to credit the reasoning explicitly: bumping Hutbit to 45 and dropping the guard would have been strictly worse, because Hutbit would then steal even a correctly named Robi S9 that happens to carry 0x02AC.

Merging this to dev now so you have a build to test against. I am landing four follow-ups on top rather than blocking your PR on them:

1. **Narrowing the OEM claim with d618.** Hutbit at 35 outranks MGB at 30 and the OEM branch has no name gate, so a genuine MGB Swan or Icomon carrying 0x02AC would now be claimed by Hutbit. The Hutbit is itself an MGB-family device, so that is a real risk rather than a theoretical one. Requiring d618 alongside ffb0 fails closed and costs nothing on your capture. Being honest about its limits: d618 rides in the same AD element as ffb0 on your unit and is probably a generic Lefu service, so it narrows against unrelated devices squatting on 0x02AC but likely does not discriminate within the Lefu family.

2. **The fix is dead on the noble target-MAC path.** `handler-noble-shared.ts:529` is not a re-resolve, it is the primary match for target-MAC mode, and the parsed manufacturer data is discarded at :272-281. So anyone on Windows or macOS with `scale_mac` set still loses the unit to MGB. One line fixes it.

3. **Two comment corrections.** The JSDoc says the payload is the MAC reversed, but the code does not check that and cannot: `BleDeviceInfo` has no address field. Worth stating only what is actually checked. Also 0x02AC is SIG-assigned to RTB Elektronik GmbH, not Lefu, so the Lefu firmware is squatting on someone else's company id. That is exactly why the id is a weak signal and the service gate carries the weight.

4. **`registry-check.ts:102-108` invariants** pin `Robi S9 > MGB` but have no Hutbit entry, and the overlap analysis skips custom adapters. Adding `Robi S9 > Hutbit > MGB` makes a future priority edit fail CI instead of silently re-breaking this.

On the `watch.test.ts` failure: that is a local macOS artifact, no action needed. CI runs ubuntu only (matrix node 22/24/26) and it passes 10/10 here on Windows. The mechanism is plausible (`src/config/watch.ts:45-49` watches the directory and filters on filename, and FSEvents can coalesce sibling events), but it is not reproducible on CI or on my box.

I am not closing #278 on this merge. The fix only works where `manufacturerData` reaches `matches()`. I verified it is populated on esphome-proxy, mqtt-proxy scan, and noble discovery, and it will be on noble target-MAC after follow-up 2. It is absent on node-ble entirely (`discovery.ts:209`, `scan.ts:271-275`), so a SWAN Hutbit on Linux native, which is the Raspberry Pi production path, still falls to MGB. That is a separate bug this exposed and I am filing it.

Your live test on the add-on is the gate for the release, purely because nobody here owns the hardware. Ping me when you have run it.

PR #284, opened 2026-07-20
Support reading stable weight

sanglt:
Summary
Fixes premature measurement completion for Eufy/1byone scales by adding a generic opt-in weight stability gate. Some scales can emit realtime weight frames before the user has fully settled, so adapters can now require consecutive matching weight readings before the shared BLE pipeline marks a measurement complete.
Changes
Added a reusable weightStability adapter capability.
Added shared consecutive-weight tracking in the GATT reading pipeline.
Opted Eufy P2/P2 Pro and 1byone/Eufy adapters into two-sample exact weight stability.
Kept adapter-specific completion logic focused on protocol semantics, such as Eufy’s final-frame flag.
Renamed the composition-hold log from Weight stable; holding... to Reading complete; holding... to avoid confusing it with raw weight stability.
Added regression coverage for the shared stability gate and affected adapters.
Test Plan

Tests pass (npm test -- tests/ble/notification-processor.test.ts tests/ble/shared.test.ts tests/interfaces/scale-adapter-types.test.ts tests/scales/eufy-p2.test.ts tests/scales/one-byone.test.ts)

Lint clean (npx eslint ...)

TypeScript compiles (npm run build)

Tested manually (describe below)
Notes
This is opt-in, not global, so scales that only emit one final frame are not forced to wait for extra samples.

### DIFF
--- a/src/ble/notification-processor.ts
+++ b/src/ble/notification-processor.ts
@@ -1,4 +1,4 @@
-import type { ScaleReading } from '../interfaces/scale-adapter.js';
+import type { ScaleReading, WeightStabilityGate } from '../interfaces/scale-adapter.js';
 import { bleLog } from './types.js';
 
 /**
@@ -72,7 +72,7 @@ export class HoldTimer {
     this.held = reading;
     if (this.timer) return;
     bleLog.info(
-      `Weight stable; holding connection up to ` +
+      `Reading complete; holding connection up to ` +
         `${Math.round(this.holdMs / 1000)}s for body composition...`,
     );
     this.timer = setTimeout(() => {
@@ -91,5 +91,48 @@ export class HoldTimer {
       clearTimeout(this.timer);
       this.timer = null;
     }
+  }
+}
+
+/**
+ * Tracks an opt-in consecutive-weight stability rule across live readings.
+ * This is deliberately outside individual adapters so protocols can keep their
+ * own "complete" semantics while sharing the common "weight stopped moving"
+ * gate where it is useful.
+ */
+export class WeightStabilityTracker {
+  private readonly requiredSamples: number;
+  private readonly toleranceKg: number;
+  private previousWeight: number | null = null;
+  private consecutiveSamples = 0;
+
+  constructor(config: NonNullable<WeightStabilityGate['weightStability']>) {
+    this.requiredSamples = Math.max(1, Math.trunc(config.samples ?? 2));
+    this.toleranceKg = Math.max(0, config.toleranceKg ?? 0);
+  }
+
+  observe(reading: ScaleReading): boolean {
+    const weight = reading.weight;
+    if (!Number.isFinite(weight)) {
+      this.reset();
+      return false;
+    }
+
+    if (
+      this.previousWeight !== null &&
+      Math.abs(weight - this.previousWeight) <= this.toleranceKg
+    ) {
+      this.consecutiveSamples += 1;
+    } else {
+      this.consecutiveSamples = 1;
+    }
+
+    this.previousWeight = weight;
+    return this.consecutiveSamples >= this.requiredSamples;
+  }
+
+  reset(): void {
+    this.previousWeight = null;
+    this.consecutiveSamples = 0;
   }
 }
--- a/src/ble/shared.ts
+++ b/src/ble/shared.ts
@@ -9,7 +9,7 @@ import type {
 } from '../interfaces/scale-adapter.js';
 import type { WeightUnit } from '../config/schema.js';
 import { LBS_TO_KG, normalizeUuid, errMsg, bleLog } from './types.js';
-import { HistoryBuffer, HoldTimer } from './notification-processor.js';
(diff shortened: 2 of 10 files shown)


KristianP26 (changes requested):
Thanks for this, and for the tests. You are fixing a real bug: the adapters completed on the first frame the scale flagged as final, and on these units that final flag can still carry a weight that is drifting, so we occasionally exported an unsettled value. Waiting for the weight to repeat is the right instinct. I read it closely against the pipeline and verified each point in the code, and there are two things I need changed before it can land, because as written it trades an occasional-wrong-value bug for an occasional-no-value bug, which is worse.

## 1. The stronger gate has no safety net, so it can now complete with nothing (blocking)

Before this PR, `isComplete` had exactly one job and always had an escape: the first final frame resolved the read. After it, completion requires two consecutive frames whose raw weight is identical (for Eufy the confirming frame must also be `isFinal`; for 1byone any two equal frames). Neither `EufyP2Adapter` nor `OneByoneAdapter` defines `completionHoldMs` or `isFinal`, so in `waitForRawReading` (`src/ble/shared.ts:419`) `isComplete` is the *only* path that resolves the promise. There is no hold timer and no timeout-to-best-effort. If the two-equal-frames condition is never met and the scale then disconnects, the read rejects at `shared.ts` with "Scale disconnected before reading completed."

The realistic failure is a scale whose final-flagged weight settles monotonically through distinct values (80.1 -> 80.0 -> 79.9, each flagged final) and then drops the link. `stableRawWeight` is `null` at every step, so nothing completes, where the old code would have returned 80.1. That monotonic-settle shape is essentially the same behaviour this PR exists to handle, so it is not hypothetical. A single-run invocation (not continuous) then exports nothing at all.

What I would accept, any of:
- Keep the stability gate as the *preferred* completion, but resolve the last plausible reading on disconnect/timeout instead of rejecting (a best-effort fallback), or
- Add `completionHoldMs` to these adapters and have the hold window resolve the last-seen stable candidate, or
- Keep the old completion as an OR fallback (see point 2, which folds into this cleanly).

Whichever route, please add a test for "final frames arrive with distinct monotonically-decreasing weights, then disconnect" and assert we still export the last one rather than nothing.

## 2. Restore an impedance floor for the Eufy BIA path (blocking, small)

The old Eufy `isComplete` required `impedance > 200` for a GATT reading. The new one drops the impedance term entirely and completes on weight stability alone. For impedance exactly 0 that is fine (the old code also completed there, weight-only, no BIA), but the new code will now also complete on a low *nonzero* transient impedance in the (0, 200] range, and `computeMetrics` will then compute body fat from that bad value (`eufy-p2.ts` computes BIA whenever `impedance > 0`). So please gate completion so a low transient impedance does not resolve, e.g. complete on weight-stable AND (impedance === 0 OR impedance > 200). That keeps your weight-only case working while not feeding garbage into BIA.

## 3. Non-blocking

- The GATT-vs-broadcast distinction via `reading === this.lastGattReading` works today only because the pipeline passes the same object reference into `isComplete` with no clone, and because `normalizesWeight = true` skips the lbs mutation at `shared.ts:402`. Both are true now, but it is an invisible dependency. Please add a one-line comment on `isComplete` noting it, or key off an explicit per-reading flag instead of object identity.
- 1byone's gate has no "measurement locked" bit to require (Eufy has `data[12]`), so two equal frames during a transient plateau can complete on a non-settled weight. Still better than the old first-frame behaviour, so not blocking, just flagging the asymmetry.
- After this change both call sites use `this.parseGattWeight`, leaving the exported `parseWeightNotification` (`eufy-p2.ts:250`) used only by tests. If you keep it exported, a comment that it bypasses stability tracking would prevent a future caller getting first-frame semantics by accident; otherwise fold it in.

Everything here is fixable and the direction is right. Land points 1 and 2 with a regression test for the settle-then-disconnect case and I will merge it to `dev`. Since none of us can test every one of these units, I would also gate the release (not the merge) on a live confirmation from someone with a P2 or a T914x that a normal weigh-in still completes.

KristianP26:
Follow-up: I have opened #285, which takes your stability idea and routes it through the existing completion-hold path (`completionHoldMs` + `isFinal`) instead of a hard two-equal-frames gate. That keeps the fix you wanted (never resolve on a still-drifting weight) while removing the never-completes case: if the weight never settles, shared.ts resolves the last held reading on hold-timeout or disconnect, so a reading is never lost, and it also keeps the Eufy impedance floor. Tests cover the settle, lone-frame, and reconnect paths; full suite green.

Your report and approach are what drove it, so thank you. I would still value your eyes on #285, and if you have a P2 or T914x handy a live confirmation there would be ideal. I will leave this PR to you to close once you have looked, rather than closing it myself.

sanglt:
Thanks for the review. I think your version is safer than my generic hard gate. That matters because my generic implementation currently blocks completion until the stability gate passes. If a scale emits only one settled-looking frame and then disconnects, we could recreate the “never completes / rejects” problem.

PR #332, opened 2026-08-20
feat(qn): decode extended-dialect b4/b1 result frames so weight reaches the exporters (#235)

hedoric:
Closes the last gap in #235. Auth (E1), the proto=0xff dialect fix and the a2 06 01 1e 23 ea trigger all shipped in v1.23.0, and my GE CS 10 G runs the whole handshake correctly — but on a completed weigh-in it never streams 0x10. It sends a burst of 0xb4/0xb1 frames that the adapter dropped at the ignoring frame opcode=… branch, so the MQTT exporter had nothing to publish even though the healthcheck passed and the session looked healthy.

What this does. Decodes the 0xb4 consolidated result frame (weight = uint16LE(bytes[11:13]) / 100), with 0xb1 03 01 (weight = uint16LE(bytes[5:7]) / 100) as a fallback when no 0xb4 arrives, and returns a ScaleReading so a reading completes and publishes. The decode is:

gated on isExtendedLongFrame — no other QN dialect is touched;
checksum-validated — the standard QN trailing sum guards against a truncated/mis-framed notification being read as a weight;
emitted once per session — the scale repeats 0xb4 ~3× then sends the 0xb1 records for the same weigh-in, so a session flag suppresses the repeats.

Verified. Weight and BMI are hardware-verified against the scale's own display: 75.20 kg, BMI 20.2 (from the 0xb1 03 03 tail), cross-checked as 75.20 / 1.93² = 20.19. Both of my connects in the attached log decode to 75.20 kg and every frame's trailing checksum validates. Tests are built from those real frames; vitest, tsc, eslint and prettier are all green.

Deliberately out of scope: impedance / body composition. The 0xb4/0xb1 frames carry a proprietary multi-frequency segmental sweep (~2,300–3,050 raw units per channel), not the single ~500 Ω whole-body value computeBiaFat expects — it divides height² / impedance, and feeding a raw channel in produces ~57 % fat. So this PR emits the reading weight-only (impedance: 0), which routes body composition through the same profile-based estimate the broadcast-only scales already use. Weight and BMI are exact; that's the part this decode is sure of. I'm happy to follow up on calibrating the impedance channels against my display values (fat 10.5 %, water 64.5 %, muscle 85 %, bone 4.5 %, visceral 3 at 75.20 kg / 193 cm) in a separate PR if useful.

### DIFF
--- a/src/scales/qn-scale.ts
+++ b/src/scales/qn-scale.ts
@@ -176,6 +176,42 @@ const EXTENDED_MEASUREMENT_TRIGGER = [0xa2, 0x06, 0x01, 0x1e, 0x23, 0xea];
 const TRIGGER_REPEATS = 2;
 const TRIGGER_GAP_MS = 150;
 
+/**
+ * Completed-weigh-in result frames on the extended dialect (#235).
+ *
+ * The 20-byte GE CS 10 G / "Fit Plus" does NOT stream 0x10 live frames after a
+ * full body-composition weigh-in. Once the impedance sweep finishes it sends a
+ * burst of result frames the adapter had been dropping at the ignore branch, so
+ * the handshake succeeded end to end yet nothing ever reached the exporters:
+ *
+ *   0xB4 .. 04 01 : consolidated result, 44 bytes. Authoritative final weight.
+ *       [7-10]  measurement timestamp, LE uint32 (scale 2000-epoch)
+ *       [11-12] final weight, LE uint16, /100 kg   (matches the scale display)
+ *       [13..]  multi-frequency / segmental impedance channels, LE uint16 each
+ *   0xB1 .. 03 01 : first multi-part record, 44 bytes. Fallback weight source.
+ *       [5-6]   in-progress weight, LE uint16, /100 kg
+ *       [7..]   impedance channels
+ *
+ * @hedoric hardware-verified both against the scale's own display: weight
+ * 75.20 kg, and BMI 20.2 in the 0xB1 03 03 tail, cross-checked as
+ * 75.20 / 1.93^2 = 20.19. Every frame carries the standard QN trailing sum
+ * checksum. The 0xB4 value is the scale's final averaged weight; the 0xB1 03 01
+ * sample reads a few hundred grams high, so 0xB4 is preferred and 0xB1 03 01 is
+ * only used when no 0xB4 arrives.
+ *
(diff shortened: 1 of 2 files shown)


KristianP26 (approved):
Approving. I re-derived the parts that decide whether this is safe rather than taking the description, and every one of them holds.

**The decode.** `b4 2c 04 01 ...` is 44 bytes with byte[1] = 0x2c as its own length, and `readUInt16LE(11)` over that frame is 0x1D60 = 7520, so 75.20 kg, which is the number on your display. The `b1 2c 03 01` fallback reads 0x1D65 = 7525 at [5], the few hundred grams high you described. The trailing sum over both frames matches the last byte, so the checksum gate is real and not decorative.

**The gate.** `isExtendedLongFrame` is only ever set from a 0x12 frame of 20 bytes or more, so no classic, no 18 byte and no 19 byte QN variant can reach this branch. Your test that feeds the same bytes to an adapter that never saw the extended 0x12 is the one that proves it, and it is the test I would have asked for.

**The reading actually completes.** This was the part I most wanted to check, because a weight-only reading is worthless if it cannot resolve. `isComplete()` takes its `reading.impedance === 0` branch and returns `weight > 0`, so the reading resolves and publishes, and `computeMetrics` routes impedance 0 to the Deurenberg estimate rather than through `computeBiaFat`. Nothing fabricates a BIA number from an uncalibrated channel.

**Not forwarding the sweep is the right call.** Raw channels in the 2,300 to 3,050 range fed to a `height² / impedance` estimator produce exactly the nonsense you measured, and shipping a wrong body-fat number is worse than shipping none. Weight and BMI are what the capture proves, and that is what this emits.

One assumption I want on the record rather than hidden: the 0xB4-before-0xB1 ordering. If a unit ever opened its burst with `b1 03 01`, the session would resolve on the less accurate value. Both of your connects have 0xB4 first, and deferring the 0xB1 would risk losing the reading entirely on a unit that emits no 0xB4, so preferring 0xB4 with 0xB1 as a fallback is the right trade. If a future capture shows the other order, the fix is a short hold rather than a re-decode.

Locally on this branch: 2175 tests across 130 files, `tsc --noEmit`, eslint and prettier all clean.

Yes to the impedance follow-up. With your display values at a known weight and height you have the one thing the project has never had for this family: a calibration point. Even a per-channel scaling that lands body fat within a couple of points would be worth more than the profile estimate.

Issue #336, opened 2026-08-22
ES-CS20M (QN classic dialect): sessions that miss the 0x12 scale-info frame never stream a weight, and qn_protocol_byte can't fix it because the classic branch ignores it


alasano:
### Description

My Renpho ES-CS20M speaks the classic QN dialect (15-byte 0x12 frames, proto byte 0x15, older firmware without AE00/AE02) over the mqtt-proxy transport with an ESP32-S3 and autonomous connect. Most weigh-ins failed with "Scale disconnected before reading completed" or the 60s reading timeout. Some completed normally on the same scale with the same config minutes apart.

The difference between good and bad sessions is whether the 0x12 scale-info frame arrives. When it arrives, the adapter learns proto 0x15, every later write echoes 0x15, and the weight streams. When it is lost between the scale and the host (frequent on the instant autonomous connects, I did not find where it dies), the session opens its unlock config with 0x00, answers the first 0x14 with a 0x20 time sync built on 0x00, and two seconds later the fallback reruns the handshake hardcoded to 0xff. The scale rejects both values silently, re-sends its 0x14 ready frame over and over, never sends 0x21 or any 0x10, and disconnects. The scale keeps the measurement: its advertisement's stored-measurement counter increments after every failed session, and it then wakes repeatedly to offload the backlog, which adds constant reconnect churn on top of the original failure.

Setting `ble.qn_protocol_byte` changed nothing, so I read the adapter code. The override is only read in the long-frame branch of the 0x12 handler. The classic branch takes `data[2]` unconditionally, `onConnected` resets the state to 0x00 and hardcodes 0x00 into the generated unlock config, and `runFallbackHandshake` overwrites the state with 0xff. For a classic-dialect scale the setting is a no-op in every code path, including the lost-0x12 path where it is needed.

I patched the adapter so the forced byte seeds the session state and survives the classic branch and the fallback, set `qn_protocol_byte: 21`, and the same scale now completes weigh-ins whether or not the 0x12 arrives. PR follows this issue.

While debugging this I also hit a separate listener leak in the mqtt-proxy handler that made the failures compound over time, that one is #338.

### Steps to Reproduce

1. ES-CS20M on `handler: mqtt-proxy` with an ESP32 and `auto_connect` (default), `ble.scale_mac` set
2. Weigh in repeatedly with debug logs on
[...]

PR #337, opened 2026-08-22
fix(qn): apply the protocol byte override to the classic dialect and the pre-0x12 phases

alasano:
Full diagnosis in #336. Short version here.

`ble.qn_protocol_byte` was only read in the long-frame branch of the 0x12 handler. The classic branch took `data[2]` unconditionally, `onConnected` reset the session state to 0x00 and hardcoded 0x00 into the generated unlock config, and `runFallbackHandshake` overwrote the state with 0xff. On a classic-dialect scale the setting was a no-op in every path. This matters on proxy transports where the 0x12 can be lost in transit, because those sessions open on 0x00 and fall back to 0xff, and a scale that wants its own byte (my ES-CS20M reports 0x15) rejects both silently, keeps re-sending 0x14, never streams, and caches the weigh-in as a stored record. Whether a weigh-in worked depended on whether the 0x12 survived, so the failure looked intermittent.

The forced byte now seeds the session state in `onConnected`, flows into the generated unlock config with its checksum recomputed, overrides the classic branch of the 0x12 handler, and is kept by the fallback, whose log line now prints the byte it chose instead of a hardcoded `proto=0xFF`. The fixed-payload `[13 09 00 01 01 02]` unlock is untouched on purpose. Its trailing byte is not a computable checksum, and scales verified working received it with 0x00, including mine.

Verified on hardware: with `qn_protocol_byte: 21`, sessions that miss the 0x12 complete the same as sessions that catch it.

Two tests added, one for the classic-dialect override and one driving a session with no 0x12 at all and asserting every protocol-bearing write carries the forced byte. Full suite passing except the pre-existing TZ-sensitive BF720 date-of-birth test, tsc/eslint/prettier clean.

### DIFF
--- a/docs/guide/configuration.md
+++ b/docs/guide/configuration.md
@@ -62,18 +62,18 @@ ble:
   # qn_protocol_byte: 0
 ```
 
-| Field                 | Required                    | Default        | Description                                                                                                                                                                                          |
-| --------------------- | --------------------------- | -------------- | ---------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------- |
-| `scale_mac`           | Recommended                 | Auto-discovery | MAC address, or a CoreBluetooth UUID on macOS (bare 32-hex as the wizard writes it, or the dashed form). Prevents connecting to a neighbor's scale.                                                  |
-| `bind_key`            | Xiaomi S800 only            | (none)         | 32-char hex per-device MiBeacon key from the Mi cloud (extract with the community Xiaomi-cloud-tokens-extractor). Decrypts only the device's own FE95 broadcast. Keep it secret; it is a credential. |
-| `handler`             | No                          | `auto`         | Transport: `auto` (local radio), `mqtt-proxy` (ESP32 over MQTT), `esphome-proxy` (ESPHome Native API). See below.                                                                                    |
-| `noble_driver`        | No                          | OS default     | `abandonware` or `stoprocent`. Overrides the default BLE driver. Only applies when `handler: auto`.                                                                                                  |
-| `adapter`             | No                          | System default | Linux only. Select a specific Bluetooth adapter (e.g., `hci0`, `hci1`). See below.                                                                                                                   |
-| `force_scale_adapter` | No                          | Auto-detect    | Name of the scale protocol adapter to use, bypassing auto-detection. Requires `scale_mac`. See below.                                                                                                |
-| `session_timeout_sec` | No                          | `120`          | Seconds one GATT session may wait for a complete reading (5 to 600). Native BLE handlers only; ignored on `mqtt-proxy` and `esphome-proxy`. See below.                                               |
-| `qn_protocol_byte`    | No                          | Auto           | QN-family scales only. Protocol byte the handshake echoes back to the scale (0 to 255). Set it only when a QN scale runs the whole handshake and then reports nothing. See below.                    |
-| `mqtt_proxy`          | If `handler: mqtt-proxy`    | (none)         | MQTT proxy connection (`broker_url`, `device_id`, `topic_prefix`, `username`, `password`, `auto_connect`, `embedded_broker_*`). See [ESP32 BLE Proxy](./esp32-proxy).                                |
-| `esphome_proxy`       | If `handler: esphome-proxy` | (none)         | ESPHome Native API connection (`host`, `port`, `encryption_key` or `password`, `client_info`). See [ESPHome Bluetooth Proxy](./esphome-proxy).                                                       |
+| Field                 | Required                    | Default        | Description                                                                                                                                                                                                                                        |
+| --------------------- | --------------------------- | -------------- | -------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------- |
+| `scale_mac`           | Recommended                 | Auto-discovery | MAC address, or a CoreBluetooth UUID on macOS (bare 32-hex as the wizard writes it, or the dashed form). Prevents connecting to a neighbor's scale.                                                                                                |
+| `bind_key`            | Xiaomi S800 only            | (none)         | 32-char hex per-device MiBeacon key from the Mi cloud (extract with the community Xiaomi-cloud-tokens-extractor). Decrypts only the device's own FE95 broadcast. Keep it secret; it is a credential.                                               |
+| `handler`             | No                          | `auto`         | Transport: `auto` (local radio), `mqtt-proxy` (ESP32 over MQTT), `esphome-proxy` (ESPHome Native API). See below.                                                                                                                                  |
+| `noble_driver`        | No                          | OS default     | `abandonware` or `stoprocent`. Overrides the default BLE driver. Only applies when `handler: auto`.                                                                                                                                                |
+| `adapter`             | No                          | System default | Linux only. Select a specific Bluetooth adapter (e.g., `hci0`, `hci1`). See below.                                                                                                                                                                 |
+| `force_scale_adapter` | No                          | Auto-detect    | Name of the scale protocol adapter to use, bypassing auto-detection. Requires `scale_mac`. See below.                                                                                                                                              |
+| `session_timeout_sec` | No                          | `120`          | Seconds one GATT session may wait for a complete reading (5 to 600). Native BLE handlers only; ignored on `mqtt-proxy` and `esphome-proxy`. See below.                                                                                             |
+| `qn_protocol_byte`    | No                          | Auto           | QN-family scales only. Protocol byte the handshake echoes back to the scale (0 to 255). Set it when a QN scale runs the whole handshake and then reports nothing, or when its scale-info frame is lost in transit on a proxy transport. See below. |
+| `mqtt_proxy`          | If `handler: mqtt-proxy`    | (none)         | MQTT proxy connection (`broker_url`, `device_id`, `topic_prefix`, `username`, `password`, `auto_connect`, `embedded_broker_*`). See [ESP32 BLE Proxy](./esp32-proxy).                                                                              |
+| `esphome_proxy`       | If `handler: esphome-proxy` | (none)         | ESPHome Native API connection (`host`, `port`, `encryption_key` or `password`, `client_info`). See [ESPHome Bluetooth Proxy](./esphome-proxy).                                                                                                     |
 
 ::: warning Forcing a scale adapter
 `force_scale_adapter` is an escape hatch for when auto-detection routes your scale to the wrong protocol adapter, which happens with rebadged OEM hardware that shares a vendor service with another brand.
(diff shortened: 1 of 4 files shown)


KristianP26 (changes requested):
The PR extends `ble.qn_protocol_byte` from the long-frame branch of the 0x12 handler to the whole session: it seeds `seenProtocolType` in `onConnected`, feeds it into the legacy unlock config, overrides the classic-dialect branch, and keeps it through the no-0x12 fallback. The diagnosis in #336 is sound, the checksum recompute is correct, and the decision to leave the fixed-payload `[13 09 00 01 01 02]` unlock alone is right. Two things need work: the unlock config's byte[2] is now decided by a documented BLE race rather than by config, so the unset-override path is no longer deterministic; and mutation testing shows two of the four behaviour hunks (the `onConnected` seed and the unlock byte) are not exercised by any test, including the new one whose name claims to cover them. Full suite passes on the PR head: 131 files, 2217 tests, tsc/eslint/prettier clean.

## Findings

### 1. Major: Unlock config byte[2] becomes timing-dependent when no override is set
`src/scales/qn-scale.ts:495`

`onConnected` builds the legacy unlock config from `this.seenProtocolType`, but it does so *after* `await this.ensureAe02Subscribed()` (qn-scale.ts:483). The method's own doc comment states that on Linux node-ble the FFF1 CCCD subscription runs in parallel with `onConnected` and the scale may send 0x12 before it finishes; src/ble/shared.ts confirms the ordering (the notify-subscribe loop at shared.ts:322-326 completes, then `await startInit()` at shared.ts:353 calls `adapter.onConnected(ctx)` at shared.ts:235). So a 0x12 arriving during the AE02 subscribe runs `parseNotification`, sets `seenProtocolType` (qn-scale.ts:799/805), and the unlock built moments later inherits it instead of the 0x00 it has always carried.

The PR description asserts the default path is unchanged; it is not. This also inverts the invariant the file documents: the handlers were required not to depend on state set in `onConnected`, and now `onConnected` depends on state the handlers set.

Direction of harm is arguable (the scale's own byte may well be the better one), but the whole premise of the surrounding comment block at qn-scale.ts:762-778 is that the byte a scale *reports* is not necessarily the byte its firmware *accepts*, and a wrong choice fails silently. Substituting a reported byte into a previously fixed frame on a timing coin-flip is the exact hazard that comment warns about.

**What goes wrong:** A classic or 19-byte QN scale without AE00, on Linux native BLE, with `qn_protocol_byte` unset. The 0x12 lands while the AE02 subscribe is still in flight and then rejects. I reproduced this with a probe test on the PR head: the unlock config written is `13 09 ff 01 10 00 00 00 2c`. With the pre-PR hardcoded 0x00 at index 2, the identical scenario yields `13 09 00 01 10 00 00 00 2d`. Same scale, same config, different frame from one session to the next depending on radio timing.

**Suggested fix:** Use `this.forcedProtocolType ?? 0x00` at index 2 instead of `this.seenProtocolType`. That is identical for every override case the PR targets (the override is set before `onConnected` runs, via `configure`), and it keeps the deterministic 0x00 when the setting is unset. The `config[8]` recompute needs no change.

### 2. Major: New test does not exercise the unlock config or the onConnected seed it claims to cover
`tests/scales/qn-scale.test.ts:1419`

The test is named "opens with the override before 0x12 ..." and its comment says "the session must then still open the unlock config with the forced byte", but its mock context uses `subscribe: async () => {}`, which resolves. That sets `hasAe00 = true`, so the `if (!hasAe02)` guard at qn-scale.ts:485 skips the legacy unlock entirely. Every `0x13 ... w[4] === 0x10` write the test collects comes from `handleScaleInfo` inside `runFallbackHandshake`, not from the unlock.

Mutation testing on the PR head, one hunk at a time:
- qn-scale.ts:805 (classic branch) reverted to `data[2]` -> 1 test fails. Covered.
- qn-scale.ts:529 (fallback) reverted to `0xff` -> 1 test fails. Covered.
- qn-scale.ts:495 (unlock byte) reverted to `0x00` -> 106/106 QN tests pass, full suite 2217/2217 passes. NOT covered.
- qn-scale.ts:453 (`onConnected` seed) reverted to `0x00` -> 106/106 pass, full suite passes. NOT covered.

The seed matters more than the tests suggest. In the reporter's failing log in #336 the 0x14 ready frame arrives before the 2s fallback fires, so `handleReady` builds `20 08 00 d7 8e 1b 32 da` from `onConnected`'s reset. That is one of the two wrong writes the issue is about, and nothing in the suite would catch its return.

**What goes wrong:** Someone later re-hardcodes `0x00` in either `onConnected`'s reset or the unlock config while refactoring. The full 2217-test suite stays green, and the reporter's exact bug ships again: a no-AE00 classic scale gets `13 09 00 ...` and `20 08 00 ...` despite `qn_protocol_byte: 21` being set.

**Suggested fix:** Two additions, both cheap. (1) The existing test at tests/scales/qn-scale.test.ts:535 already builds a context whose `subscribe` rejects, so the unlock path runs there; add `adapter.configure({ qnProtocolByte: 0x15 })` and assert `config![2] === 0x15` alongside the existing unit-flag and checksum assertions. (2) In the new fallback test, deliver the 0x14 ready frame (the reporter's `14 0b 15 00 ...`) before advancing to 2000ms and assert the resulting `0x20` time sync carries 0x15, which pins the `onConnected` seed rather than the fallback assignment.

### 3. Minor: Public docs still describe the old, narrower use of qn_protocol_byte
`docs/guide/configuration.md:74`

The PR broadened the schema JSDoc (src/config/schema.ts:135-141) to cover a scale-info frame lost in transit on proxy transports, but docs/guide/configuration.md was not touched. The table row still reads "Set it only when a QN scale runs the whole handshake and then reports nothing", and the tip at lines 94-113 says the same in prose.

Two further points in that tip are now actively misleading for the case this PR fixes. It tells users to "try the other value" and shows `qn_protocol_byte: 0 # or 255`, framing the setting as a binary choice, while the hardware this PR was verified on needs 21. And it tells users to confirm the setting via the line `QN: scale info (19B, dialect=es26m), factor=10, proto=0xff`, which is precisely the line that never prints when the 0x12 is lost. The PR's improved fallback line is the one those users will see.

**What goes wrong:** An ES-CS20M or Arboleaf owner on esphome-proxy or mqtt-proxy reads the tip, tries 0 and 255, sees no change because their scale wants its own byte, and concludes the setting does not apply to them, which is the exact dead end #336 describes.

**Suggested fix:** Update the row at line 74 to match the new schema wording, and in the tip: mention that a scale may need its own byte (the full 0 to 255 range is accepted, and an ES-CS20M reporting 21 is a real example), and add the fallback log line `QN: fallback: no 0x12 received, running handshake with proto=0x15` next to the existing scale-info line as the confirmation to look for when the scale-info frame does not arrive.

### 4. Minor: Scale-info debug line no longer distinguishes a reported byte from a forced one on the classic dialect
`src/scales/qn-scale.ts:816`

The log prints `this.seenProtocolType`, which after this PR is the forced byte whenever the override is set. On the classic dialect that byte used to always be what the scale reported, so the line was usable as ground truth during triage. It no longer is. The same ambiguity already existed for the long-frame branch before this PR, so this extends an existing wart rather than creating one, but the classic dialect is where reporters are now being asked to set the override.

**What goes wrong:** A reporter sets `qn_protocol_byte: 21` on a classic scale that actually reports 0xab, pastes `QN: scale info (15B, dialect=classic), factor=100, proto=0x15` into an issue, and triage records 0x15 as that model's reported protocol byte. The real value is never learned.

**Suggested fix:** When `forcedProtocolType` is set and differs from the byte the frame carried, append it, e.g. `proto=0x15 (forced, scale reported 0xab)`. Both values are in scope at that point in the branch.

### 5. Minor: HA add-on users cannot set qn_protocol_byte at all
`ble-scale-sync-addon/config.yaml:68`

The add-on options schema has no `qn_protocol_byte` entry, and ble-scale-sync-addon/run.sh writes only `scale_mac`, `adapter` and `force_scale_adapter` into the generated `ble:` block (run.sh:139-145). This is pre-existing from the commit that introduced the setting, not a regression in this PR, but it is worth naming here because this PR's stated audience is proxy-transport users, which overlaps heavily with the add-on crowd.

**What goes wrong:** An add-on user on esphome-proxy hits exactly the #336 symptom, reads the docs, and has nowhere to put the setting: `run.sh` regenerates config.yaml on every start, so hand-editing it does not survive a restart.

**Suggested fix:** Follow-up, not a blocker for this PR: add `qn_protocol_byte: int(0,255)?` to the add-on schema and emit it from run.sh alongside the other ble keys, then extend the `ble:` block guard at run.sh:139 to include it.

## Worth calling out

- **Checksum recompute is correct and the fixed-payload unlock was rightly left alone** (src/scales/qn-scale.ts:503) `config[8]` is initialised to 0x00 and the reduce sums indices 0 through 8, so the recompute does cover the changed byte at index 2. Verified against the reporter's live capture in #336: 0x13+0x09+0x15+0x02+0x10 = 0x43, matching `13 09 15 02 10 00 00 00 43`, and 0x2e for the 0x00 form.

Leaving `[13 09 00 01 01 02]` untouched is the right call and the stated reason checks out: the sum of its first five bytes is 0x1e, not 0x02, so the trailing byte is an opaque literal rather than a checksum and must not be regenerated. Printing the chosen byte in the fallback log instead of a hardcoded `proto=0xFF` is also a good change, since that line is the only diagnostic a user gets in the no-0x12 case.

## Questions for you

- In a working session on your ES-CS20M (the one where the 0x12 does arrive), does the untrimmed debug log show `QN: scale info (15B, dialect=classic)` before or after the two `QN write: [13 09 ...]` unlock lines? That pins down on real hardware whether the 0x12-versus-AE02-subscribe race in finding 1 actually fires on your unit.

- In that same working session, what byte[2] does the unlock config `13 09 ?? 02 10 ...` carry with your patch applied and `qn_protocol_byte: 21` set? Confirms the unlock hunk is doing what the PR intends on the live path.

alasano:
Changed the unlock byte to `forcedProtocolType ?? 0x00`. It's the same for every override case since `configure()` runs before `onConnected()`, and it's back to a fixed `0x00` when nothing is set.

The two tests you asked for are in. The no-AE00 unlock path asserts byte[2] is the forced value and `0x00` when unset, and the fallback test feeds a `0x14` ready frame before the 2 s timer so the `0x20` it emits pins the `onConnected` seed. I added a third on top because your mutation check has a variant those two don't catch. Reverting the line to the old `this.seenProtocolType` rather than to a literal `0x00` leaves the whole suite green, so the third test holds the AE02 subscribe open, delivers a 0x12 carrying `0xab` while it's pending, then rejects the subscribe and asserts the no-override unlock still opens with `0x00`. On the old line it comes out `0xab`.

The classic-dialect scale-info log now appends `(forced, scale reported 0x..)` when the override differs from what the frame carried, and it reports the raw `data[2]` for every dialect, so the 18-byte variant logs the `0xff` it actually sent even though the handshake opens with `0x00`. Docs updated as suggested, with the ES-CS20M reporting 21 as the real example and the fallback line as the confirmation to look for when the scale-info frame never arrives.

Your two questions, from fresh sessions on my ES-CS20M with `qn_protocol_byte: 21` set. The 0x12 never arrives on this unit over mqtt-proxy. Five sessions back to back today, all `fallback: no 0x12 received`, including the one that produced a reading. So I can't answer the before/after ordering on this hardware, but it also means the race from finding 1 can't fire on this transport, the unlock byte here is decided entirely by config.

Byte[2] on the unlock is `0x15`, the writes are `13 09 15 02 10 00 00 00 43`. One working session, trimmed:

```
QN write: [13 09 00 01 01 02]
QN write: [13 09 15 02 10 00 00 00 43]
Subscribed to notifications. Step on the scale.
...
QN: fallback: no 0x12 received, running handshake with proto=0x15
QN write: [13 09 15 02 10 00 00 00 43]
QN: fallback: sending history + start
...
Notification 0000ffe1...: [10 0b 15 23 96 01 01 fe 01 f0 da]
Reading complete: 91.10 kg / 510 Ohm
```

KristianP26 (approved):
Both points from the last round are properly closed, and I verified the tests rather than reading them.

Mutation check on the merged tree, one hunk at a time, running `tests/scales/qn-scale.test.ts` (108 tests):

- `onConnected` seed back to `0x00`: 1 failed
- unlock byte[2] back to `this.seenProtocolType`: 1 failed, `expected 171 to be +0`, which is your third test catching the `0xab` the racing 0x12 seeds
- classic branch back to `data[2]`: 1 failed
- fallback back to `0xff`: 1 failed

So all four behaviour hunks are pinned, including the two that were uncovered last time. The unlock config is deterministic again (`forcedProtocolType ?? 0x00`), and reporting the raw `data[2]` in the scale-info line is the right call: the 18-byte variant now logs the `0xff` it actually sent while the handshake still opens on `0x00`, which is exactly the triage line that was missing.

Full suite on `dev` plus this branch: 131 files, 2228 tests, tsc, eslint and prettier all clean. CI green on 22/24/26.

Note for `#351`: this touches `docs/guide/configuration.md`, so that branch will want this in its rebase base.

Issue #338, opened 2026-08-22
mqtt-proxy: notify listeners leak on timed-out GATT sessions until one scale frame is processed hundreds of times


alasano:
### Description

I run the mqtt-proxy transport with an ESP32-S3 and autonomous connect. After the container had churned through a few hours of failed GATT sessions (the scale problem in #336), the debug log started showing every single notification frame processed several times, and eventually hundreds of times. One 11-byte ready frame from the scale produced hundreds of identical `Notification` / `QN RAW` / `ready frame, sending time sync + profile` triplets with timestamps inside the same 10 ms. Each processing queued more writes toward the ESP32, so the scale got buried in duplicate handshake traffic and live weigh-ins died. A container restart resets it, and the duplication then builds back up as sessions fail.

The giveaway in the log is that "sending time sync + profile" repeats hundreds of times while the actual 0x20 write appears once. `handleReady()` gates on `timeSyncSent`, so this is one adapter being invoked by many listeners, and the adapter is a singleton, so the duplication has to be on the MQTT message-listener side.

Reading the handler code, I found three places where a listener stays on the persistent MQTT client with nothing left that can remove it:

1. A session that hits the 60s reading timeout in the watcher. `withTimeout` races the reading promise but does not cancel it, and the watcher's `finally` calls `device.cleanup()`, which removes only the disconnected-topic relay. The notify listener belongs to the still-pending `waitForRawReading`, whose only cleanup path is the disconnect callback, and that callback can no longer fire because the relay was just removed. One leaked notify listener per timed-out session, forever.
2. `MqttBleChar.subscribe()` attaches its listener before awaiting the broker subscribe and publish. If either await rejects, the unsubscriber is never returned, and the listener leaks.
3. `mqttGattConnect()`'s response handler removes itself only when a connected or error message arrives. A connect timeout leaves it attached.

The first one is the one that compounds in normal operation, because a scale that advertises while nobody stands on it (mine does, after failed weigh-ins it wakes repeatedly to offload cached readings) generates timed-out sessions all day.

PR follows this issue.

### Steps to Reproduce

1. mqtt-proxy with `auto_connect` and `ble.scale_mac` set, continuous mode
2. Let the ESP32 autonomously connect to a scale that then sends no reading, so the session ends in `GATT reading timeout` (a QN scale that advertises with cached data does this on its own)
3. Repeat. After N timed-out sessions, each notification frame in the next session is processed N+1 times
[...]

Issue #138, opened 2026-04-26
support for scale BA 1600 by ADE


sttehh:
### Description

I would like to have support for scale BA 1600 by ADE which is similar to recognised scale Trisa

**BLE scan:**
docker run --rm -it --network host --cap-add NET_ADMIN --cap-add NET_RAW   --group-add 112 -v /var/run/dbus:/var/run/dbus:ro  --security-opt apparmor=unconfined  -e DEBUG=true   ghcr.io/kristianp26/ble-scale-sync:latest scan
Resetting Bluetooth adapter...
Set Powered for hci0 failed with status 0x14 (Permission Denied)
Set Powered for hci0 failed with status 0x14 (Permission Denied)
Bluetooth adapter reset OK
2026-04-26 08:25:07.603 [Scan] Scanning for BLE devices... (15 seconds)

2026-04-26 08:25:07.609 [BLE:debug] BLE handler: node-ble (BlueZ D-Bus)
2026-04-26 08:25:07.758 [BLE:debug] startDiscovery failed: Discovery already in progress
2026-04-26 08:25:07.760 [BLE:debug] Discovery already active (owned by another client), continuing
2026-04-26 08:25:24.059 [Scan]   04:0D:84:E4:60:E1  Name: Hue Go
2026-04-26 08:25:24.060 [Scan]   24:E5:AA:BE:98:8E  Name: Philips Sonicare
2026-04-26 08:25:24.061 [Scan]   2C:BC:BB:38:5D:62  Name: ShellyPlus2PM-2CBCBB385D60
2026-04-26 08:25:24.061 [Scan]   42:56:39:8D:A2:BD  Name: (unknown)
2026-04-26 08:25:24.062 [Scan]   45:F7:10:E8:0D:CE  Name: (unknown)
2026-04-26 08:25:24.062 [Scan]   54:0F:2C:44:D6:07  Name: (unknown)
2026-04-26 08:25:24.062 [Scan]   54:66:8C:85:BC:00  Name: (unknown)
2026-04-26 08:25:24.063 [Scan]   59:35:B2:07:13:5A  Name: (unknown)
2026-04-26 08:25:24.064 [Scan]   5B:E8:77:66:CA:69  Name: (unknown)
2026-04-26 08:25:24.064 [Scan]   62:2C:61:43:A7:3D  Name: (unknown)
2026-04-26 08:25:24.065 [Scan]   63:B4:C6:42:DC:95  Name: (unknown)
[...]

PR #339, opened 2026-08-22
fix(mqtt-proxy): stop leaking message listeners from timed-out and failed GATT sessions

alasano:
Full diagnosis in #338. Three paths left handlers on the persistent MQTT client with nothing to remove them.

A session hitting the 60s reading timeout abandoned `waitForRawReading` mid-flight. The watcher's cleanup removed the disconnected-topic relay but not the notify listeners held by the pending promise, and removing the relay also removed the only path that could still clean them. The watcher now fires the local disconnect before dropping the relay, so the abandoned session tears itself down through its normal disconnect path. The abandoned promise gets a no-op catch first, because its rejection now actually happens and has no other consumer, and an unconsumed rejection would take the process down on current Node.

`MqttBleChar.subscribe()` attached its listener before awaiting the broker subscribe and the firmware notify command. A failure in either await meant no unsubscriber was ever returned. The listener is now removed on setup failure before the error propagates.

`mqttGattConnect()`'s response handler removed itself only when a connected or error message arrived, so a connect timeout leaked it. The handler now lives outside the promise executor and is removed in a `finally` around the wait, which also let the executor's three self-removal calls collapse into that single removal point.

Every leaked listener re-processed all later notifications, so each timed-out session compounded. After an evening of them, one scale frame was handled hundreds of times, each firing duplicate handshake writes, which buried the scale and killed live weigh-ins until a restart.

Three tests added: an autonomous session that times out must return the client to its baseline listener count, a failed `subscribe()` setup must not leave its listener behind, and a timed-out `mqttGattConnect` must remove its response handler. Full suite passing except the pre-existing TZ-sensitive BF720 date-of-birth test, tsc/eslint/prettier clean.

### DIFF
--- a/src/ble/handler-mqtt-proxy/gatt.ts
+++ b/src/ble/handler-mqtt-proxy/gatt.ts
@@ -17,13 +17,21 @@ export class MqttBleChar implements BleChar {
       if (t === topic) onData(payload);
     };
     this.client.on('message', handler);
-    await this.client.subscribeAsync(topic);
-    // Ordering is the whole point of #231: the MQTT notify subscription and the
-    // message handler are in place BEFORE we tell the firmware to enable BLE
-    // notify, so the firmware-triggered kickoff frame (QN/Renpho 0x12) always has
-    // a listener. New firmware enables notify on this command; old firmware (eager)
-    // ignores it and behaves exactly as before.
-    await this.client.publishAsync(`${this.base}/subscribe/${this.uuid}`, '');
+    try {
+      await this.client.subscribeAsync(topic);
+      // Ordering is the whole point of #231: the MQTT notify subscription and the
+      // message handler are in place BEFORE we tell the firmware to enable BLE
+      // notify, so the firmware-triggered kickoff frame (QN/Renpho 0x12) always has
+      // a listener. New firmware enables notify on this command; old firmware (eager)
+      // ignores it and behaves exactly as before.
+      await this.client.publishAsync(`${this.base}/subscribe/${this.uuid}`, '');
+    } catch (err) {
+      // A failed setup returns no unsubscriber, so the listener must not outlive
+      // this call: the client is persistent and every leaked listener re-processes
+      // all later notifications.
+      this.client.removeListener('message', handler);
+      throw err;
+    }
     return () => {
       this.client.removeListener('message', handler);
     };
@@ -104,41 +112,52 @@ export async function mqttGattConnect(
   await client.subscribeAsync(t.disconnected);
   await client.subscribeAsync(t.error);
 
-  const response = await withTimeout(
-    new Promise<{ chars: Array<{ uuid: string; properties: string[] }> }>((resolve, reject) => {
-      const handler = (topic: string, payload: Buffer) => {
-        if (topic === t.connected) {
-          let data: Record<string, unknown>;
-          try {
-            data = JSON.parse(payload.toString());
-          } catch (err) {
-            client.removeListener('message', handler);
-            reject(new Error(`Invalid connected payload from ESP32: ${err}`));
-            return;
-          }
-          // Ignore autonomous connects from ESP32 — those are handled by
-          // ReadingWatcher.handleAutonomousConnect, not mqttGattConnect (#201).
-          if (data.autonomous) return;
-          client.removeListener('message', handler);
-          resolve(data as { chars: Array<{ uuid: string; properties: string[] }> });
-        }
-        if (topic === t.error) {
-          client.removeListener('message', handler);
-          // Older firmware (and MicroPython exceptions like asyncio.TimeoutError
-          // whose str() is empty) can publish a blank payload — surface a
-          // placeholder so the host log is not a dangling "ESP32 error:".
-          const detail = payload.toString() || '(no detail)';
-          reject(new Error(`ESP32 error: ${detail}`));
-        }
-      };
-      client.on('message', handler);
-      client
-        .publishAsync(t.connect, JSON.stringify({ address, addr_type: addrType }))
-        .catch(reject);
-    }),
-    COMMAND_TIMEOUT_MS,
-    `GATT connect timeout for ${address}`,
+  let resolveResponse!: (v: { chars: Array<{ uuid: string; properties: string[] }> }) => void;
+  let rejectResponse!: (err: Error) => void;
+  const responsePromise = new Promise<{ chars: Array<{ uuid: string; properties: string[] }> }>(
+    (resolve, reject) => {
+      resolveResponse = resolve;
+      rejectResponse = reject;
+    },
   );
+  const handler = (topic: string, payload: Buffer) => {
+    if (topic === t.connected) {
+      let data: Record<string, unknown>;
+      try {
+        data = JSON.parse(payload.toString());
+      } catch (err) {
+        rejectResponse(new Error(`Invalid connected payload from ESP32: ${err}`));
+        return;
+      }
+      // Ignore autonomous connects from ESP32 — those are handled by
+      // ReadingWatcher.handleAutonomousConnect, not mqttGattConnect (#201).
+      if (data.autonomous) return;
+      resolveResponse(data as { chars: Array<{ uuid: string; properties: string[] }> });
+    }
+    if (topic === t.error) {
+      // Older firmware (and MicroPython exceptions like asyncio.TimeoutError
+      // whose str() is empty) can publish a blank payload — surface a
+      // placeholder so the host log is not a dangling "ESP32 error:".
+      const detail = payload.toString() || '(no detail)';
+      rejectResponse(new Error(`ESP32 error: ${detail}`));
+    }
+  };
+  client.on('message', handler);
+  let response: { chars: Array<{ uuid: string; properties: string[] }> };
+  try {
+    client
+      .publishAsync(t.connect, JSON.stringify({ address, addr_type: addrType }))
+      .catch(rejectResponse);
+    response = await withTimeout(
+      responsePromise,
+      COMMAND_TIMEOUT_MS,
+      `GATT connect timeout for ${address}`,
+    );
+  } finally {
+    // The handler previously removed itself only when a response arrived, so a
+    // connect timeout left it on the persistent client forever.
+    client.removeListener('message', handler);
+  }
 
   const charMap = new Map<string, BleChar>();
   for (const char of response.chars) {
--- a/src/ble/handler-mqtt-proxy/watcher.ts
+++ b/src/ble/handler-mqtt-proxy/watcher.ts
@@ -469,21 +469,31 @@ export class ReadingWatcher implements Watcher {
         entry.address,
       );
       bleLog.debug(`GATT read driven by adapter: ${gattAdapter.name} (${entry.address})`);
-      const raw = await withTimeout(
-        waitForRawReading(
-          connected.charMap,
-          device,
-          gattAdapter,
-          profile,
-          entry.address.replace(/[:-]/g, '').toUpperCase(),
-        ),
-        60_000,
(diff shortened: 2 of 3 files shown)


KristianP26 (commented):
Three real listener leaks on the persistent MQTT client are correctly closed: `MqttBleChar.subscribe()` now removes its handler when broker setup fails, `mqttGattConnect()` moves its response handler out of the promise executor and removes it in a `finally` (previously a connect timeout leaked it forever), and both watcher GATT paths now fire the local disconnect so an abandoned `waitForRawReading` tears itself down. I verified the `fireDisconnect()`-on-an-already-resolved-session concern is safe in every direction: the disconnect callback short-circuits on `resolved` (src/ble/shared.ts:534), `resolve`/`reject` on a settled promise are no-ops, and `adapter.onSessionEnd()` is guarded by the `sessionEnded` flag (src/ble/shared.ts:200), so there is no double-resolve, no duplicate `queue.push`, and no double `onSessionEnd`. The three new tests each genuinely fail without the corresponding source change. One residual hole in the same defect class remains, in `src/ble/shared.ts` which this PR does not touch, plus two comments whose stated rationale is factually wrong.

## Findings

### 1. Major: Residual listener leak: the legacy-mode notify unsubscriber is registered only after onConnected settles
`src/ble/shared.ts:384`

The PR's premise is that firing the local disconnect makes an abandoned session release its listeners through `init.cleanup()`, which walks the `unsubscribers` array. In legacy mode that array does not yet contain the notify unsubscriber while `onConnected` is still in flight:

```ts
const [unsub] = await Promise.all([
  subscribeToChar(charMap, effectiveNotifyUuid, onNotification, adapter.name),
  startInit(),
]);
unsubscribers.push(unsub);
```

`MqttBleChar.subscribe()` attaches its `message` listener at src/ble/handler-mqtt-proxy/gatt.ts:19 *before* its own awaits, so the listener is live on the persistent client from the first tick, but `unsubscribers.push(unsub)` runs only after BOTH promises settle. If `startInit()` (that is, `adapter.onConnected`) rejects or never settles, the push never happens and `init.cleanup()` iterates an array that is missing exactly the notify listener the PR set out to remove. Twelve shipped adapters take this branch (they declare `onConnected` with no `characteristics` array): qn-scale, one-byone, inlife, renpho-es26bb, koogeek-s1, ade-a2, digoo, excelvan-cf369, exingtech-y1, hoffen, medisana-bs44x, mgb.

The multi-char branch has a narrower version of the same window at src/ble/shared.ts:324: each `unsub` is pushed right after its own await, so a subscribe that is still in flight when cleanup runs also leaks.

This is pre-existing and outside the files the PR edits, so it does not block the merge, but it is the same bug and the fix here does not close it.

**What goes wrong:** mqtt-proxy handler, 1byone/Eufy adapter (legacy branch, `onConnected` writes a unit command and a clock sync). ESP32 publishes an autonomous connect. `waitForRawReading` runs `Promise.all([subscribeToChar(fff4), init.start()])`; `MqttBleChar.subscribe` has already added its `message` listener to the persistent client. `ctx.write(charWriteUuid, unitCmd, false)` rejects because `client.publishAsync` rejects (client closing) or hangs past 60s (broker outage, mqtt.js queues QoS 0 until reconnect). `Promise.all` never resolves, so `unsubscribers` stays empty. At 60s the new `device?.fireDisconnect()` runs `init.cleanup()` over an empty array; `device?.cleanup()` removes only the disconnect relay. The fff4 notify listener stays on the persistent client -> exactly the compounding re-processing described in the PR body, one listener per failed session.

**Suggested fix:** Make `initializeAdapter` own registration instead of the call sites: track a `cleanedUp` flag alongside `sessionEnded`, expose `addUnsubscriber(u)` that calls `u()` immediately when `cleanedUp` is already true, and have `subscribeAndInit` register through it, e.g. `subscribeToChar(...).then(addUnsubscriber)` composed into the `Promise.all` rather than pushing after it. Either land it here or file a follow-up referencing #338.

### 2. Minor: pending.catch(() => {}) comment states an incorrect rationale: withTimeout already consumes the rejection
`src/ble/handler-mqtt-proxy/watcher.ts:479`

The comment says the abandoned promise's rejection "has no other consumer and must not become an unhandled rejection", and the PR body says "an unconsumed rejection would take the process down on current Node". Both are wrong. `withTimeout` (src/ble/types.ts:167) is `Promise.race([promise, timeout])`, and `Promise.race` attaches `then(resolve, reject)` to every input synchronously at call time. That marks `pending` as handled permanently; once a promise has a reaction attached it can never be reported as an unhandled rejection, even after the race has already settled with the timeout error. I confirmed this with a standalone Node repro of the exact shape (race times out, the abandoned promise is rejected afterwards, no `unhandledRejection` fires).

Same comment and same no-op at the autonomous path (line ~608, `// See the host-initiated path: the abandoned promise must stay consumed.`).

The `.catch()` itself is harmless belt and braces and I would not ask you to delete it, but the comment will teach the next maintainer that `withTimeout` drops rejections, which is the opposite of what it does.

**What goes wrong:** Not a runtime failure. A future maintainer reads this comment, concludes `withTimeout` leaves rejections unconsumed, and adds defensive `.catch(() => {})` calls (or worse, a real `.catch` that swallows a genuine error) at every other `withTimeout` call site in src/ble/.

**Suggested fix:** Reword to what it actually is, for example: `// Defensive: keep an explicit consumer on the abandoned promise so the no-op stays correct even if withTimeout stops racing it.` Drop the matching sentence from the PR description too.

### 3. Minor: Comment asserts an ordering requirement between fireDisconnect() and cleanup() that does not exist
`src/ble/handler-mqtt-proxy/watcher.ts:488`

The comment opens with "End the reading session before dropping the disconnect relay", implying `cleanup()` would break `fireDisconnect()` if it ran first. It would not. `MqttBleDevice.cleanup()` (src/ble/handler-mqtt-proxy/gatt.ts:91) only does `this.client.removeListener('message', this.handler)`; it never touches `disconnectCb`, and `fireDisconnect()` (gatt.ts:87) invokes `this.disconnectCb?.()` directly rather than through the relay. Either order works. The rest of the comment (why the disconnect must be fired at all) is accurate and valuable.

**What goes wrong:** Not a runtime failure. Someone later reorders or refactors this finally block, believes the ordering is load-bearing, and either preserves a constraint that costs them a cleaner refactor or reports a phantom regression when the order changes.

**Suggested fix:** Drop the ordering clause and keep the substance: `// End the reading session: after a timeout the abandoned waitForRawReading still holds its notification listeners on the persistent MQTT client, and only its own disconnect path removes them. ...`

### 4. Minor: Em dash reintroduced on two relocated comment lines
`src/ble/handler-mqtt-proxy/gatt.ts:132`

The `mqttGattConnect` rewrite lifts the handler out of the promise executor, so these two comments are re-added by the diff with their em dashes intact:

- line 132: `// Ignore autonomous connects from ESP32 — those are handled by`
- line 139: `// whose str() is empty) can publish a blank payload — surface a`

The text is unchanged, only moved, but the project forbids the em dash in all prose including source comments, and these lines are already being touched.

**What goes wrong:** Not a runtime failure. A later grep for the forbidden character flags these lines and they get attributed to this change, or they are copied as a template into a new comment.

**Suggested fix:** Rewrite both without the dash, for example `// Ignore autonomous connects from ESP32. Those are handled by` and `// whose str() is empty) can publish a blank payload, so surface a`.

## Worth calling out

- **Firing the disconnect also finally runs onSessionEnd() on abandoned mqtt-proxy sessions** (src/ble/handler-mqtt-proxy/watcher.ts:496) This buys more than listener removal. Before the change, a 60s reading timeout left `waitForRawReading`'s promise pending forever, so `init.cleanup()` (src/ble/shared.ts:189) never ran and `adapter.onSessionEnd()` was never called on that path. For QN Scale that means `fallbackTimer` and `storedRetryTimer` stayed armed and `this.ctx` kept pointing at the dead session (qn-scale.ts:1142 documents exactly this as the #138 defect class), so a stored-data retry could fire and write through a dead link between the timeout and the next connect. Combined with a leaked notify listener feeding the dead session's `handleNotification`, that is a very plausible amplifier for the duplicate-handshake storm reported in #338. Good instinct routing the teardown through the adapter's own disconnect path instead of bolting on ad-hoc listener bookkeeping in the watcher.

- **All three tests are genuine regression tests, verified against the pre-change code** (tests/ble/handler-mqtt-proxy.test.ts:2352) I traced each one statically against the pre-change sources:

1. Autonomous timeout test: after the autonomous connect the client holds baseline + 2 listeners (one from `MqttBleDevice`, one from `MqttBleChar.subscribe`). Without the `device?.fireDisconnect()` line, the 60s timeout only runs `device.cleanup()`, leaving the notify listener behind, so the final `toBe(baseline)` sees baseline + 1 and fails. The `advanceTimersByTimeAsync(10)` correctly flushes the microtask chain before the 60s timer is armed, and `toBeGreaterThan(baseline)` makes the test fail loudly rather than vacuously if the adapter ever stops resolving.
2. `subscribe` failure test: pre-change there is no try/catch, so the listener added at gatt.ts:19 survives the `broker down` rejection and `toHaveLength(0)` fails.
3. `mqttGattConnect` timeout test: pre-change the handler removed itself only inside the executor's connected/error branches, so a timeout leaves it attached and `toHaveLength(0)` fails. Creating the `rejects` assertion before advancing timers also keeps the rejection handled.

All three use fresh `mockClient` instances from `beforeEach` and restore real timers in `finally`, so no cross-test contamination.

## Questions for you

- Do you want the src/ble/shared.ts unsubscriber-registration guard in this PR, or should it be a follow-up issue linked to #338? It is the same leak class and the current fix does not reach it for the twelve legacy-branch onConnected adapters.

- You mention the full suite passes except a pre-existing TZ-sensitive BF720 date-of-birth test. Can you confirm CI (typecheck-and-test on 22/24/26) is green on the branch, and link or file the TZ flake separately so it does not get absorbed into this PR's history?

KristianP26:
Answering my own question about the residual leak in `src/ble/shared.ts`: I took it as a follow-up rather than asking you to widen this PR, and it is on `dev` now as 6b2138e.

For the record, since it is the same defect class you are fixing here. Legacy mode races the notify subscribe against the adapter handshake:

    const [unsub] = await Promise.all([
      subscribeToChar(...),
      startInit(),
    ]);
    unsubscribers.push(unsub);

When `startInit()` rejects, `Promise.all` rejects immediately and abandons the subscribe. The caller's `.catch` runs `init.cleanup()`, which drains an unsubscriber list the notify entry never reached, and the subscribe then resolves and installs its listener with nothing left to remove it. Native transports hide it because the link dies with the session; the proxy transports do not, which is exactly the compounding you described in #338.

The fix registers from inside the subscribe chain and routes every registration through a registrar that knows whether the session is already closed, so a subscribe settling after cleanup is unsubscribed on the spot. That also covers `ConnectionContext.subscribe`, so an adapter firing a subscribe without awaiting it cannot strand one either. Two regression tests in `tests/ble/shared.test.ts`, both verified to fail on the unpatched source.

Your PR still stands on its own and I have not touched anything it changes, so there is no rebase for you here. The remaining items from my review are the two comments whose stated rationale is wrong and the em dash on the relocated lines in `gatt.ts`. Fix those and I will merge.

On CI: I can see the checks green on the branch, so no need to chase that. Please do file the timezone-sensitive BF720 test separately if it is still biting you locally, since #348 turns out to need a timezone-pinned test of its own and the two should not collide.

KristianP26 (approved):
Re-verified on the current `dev`: merges clean, and the merged tree runs 131 files / 2224 tests green. The interaction I was worried about is fine, since the registrar landed on `dev` in 6b2138e closes the legacy-mode window from the inside while this PR closes the three watcher-side paths, and the two do not double-unsubscribe: the registrar unsubscribes only a subscribe that settles after cleanup, and `fireDisconnect()` short-circuits on an already-resolved session.

PR #341, opened 2026-08-22
feat(scales): add Salter SA00656/SA00432 (0xFFCC polled record protocol)

albert-canfield:
Adds support for Salter Bluetooth body-analyser scales, decoded from an iOS PacketLogger trace of the Salter Health app and then corrected against live hardware over a long verification session.

The scale never transmits unsolicited: no weight in the advertisement, and nothing at all over GATT until the client writes to FFC1. FFC2 and FFC3 are advertised as notify and the vendor app subscribes to both, but neither ever carries a byte. Measurements are stored in a rotating eight-slot buffer and kept until something clears them, so the adapter walks all eight slots and de-duplicates on each record's own timestamp.

Protocol (all on FFC1, write-without-response, replies notify back):

  01 <unix u32 LE>  set clock        02             read clock
  03 <i> <5 bytes>  write profile    04 <i>         read profile
  08 <i>            record status    09 <i> 00      read record i
  0a <i> 00         clear record     0b 00          keepalive

  record: <slot> 00 | <unix u32 LE> | weight, fat%, water%, muscle%,
          bone kg, BMR kcal, BMI   (u16 LE, /10 except BMR)

Three behaviours were only discoverable on hardware, and each shapes the design:

- The firmware has a single command buffer. Nine writes fired back-to-back got three answers, and the only fetch that survived was the last one written, so the slot walk is chained one fetch per reply rather than sent as a burst.
- The clear (0a) consumes a record independently of which slot was read. Paired with a fixed-slot fetch it destroyed three unread weigh-ins on a real scale, so this adapter issues no destructive write at all and relies on the buffer rotating by itself.
- The scale's clock does not agree with the host's (one unit was an hour out). Record ages are therefore derived by comparing against the scale's OWN clock, where the offset cancels.

Body composition is deliberately not taken from the scale. The six derived fields decode exactly -- verified field by field against the vendor app -- but are computed on-device from a profile the app provisions, and it had written a height of 100 cm; the scale's BMI field reported 85.7 for an 85.7 kg person as a result. Even provisioned correctly they would describe whoever the device was last told about rather than the matched user, so readings are weight-only and composition comes from the shared estimator, as with the Hutbit adapter.

Verified end to end against a SALTER-SA00656-BK: weights confirmed against the scale's own display (87.6, 85.7, 88.0, 88.8, 89.7, 7.4 kg), and a single session collecting four stored readings from slots 1, 2, 6 and 7 with correct dates.

Known limitation: normalizesWeight is inferred, not measured. Every capture was taken with the scale displaying kg; the inference rests on bone mass arriving in kilograms and BMR in kilocalories alongside the weight field. Documented in source as the first thing to check if a lb-configured unit reads 2.2x wrong.

## Changes

-

## Test Plan

- [ ] Tests pass (`npm test`)
- [ ] Lint clean (`npm run lint`)
- [ ] TypeScript compiles (`npx tsc --noEmit`)
- [ ] Tested manually (describe below)


KristianP26 (changes requested):
Adds a Salter SA00656/SA00432 adapter for a polled 0xFFCC record protocol: a 3 s keepalive plus clock read kicks off a self-chaining walk of eight record slots, one fetch per reply, and the newest record wins. The decode itself is excellent and I verified every field claim arithmetically against the captured fixtures; registration, priority (125, unique), match surface, tsc, eslint, prettier and the full test suite all check out (the one full-suite failure, tests/config/watch.test.ts, is a pre-existing Windows timing flake that passes in isolation on dev too). Two things block merge: the final commit on the branch removed the record age gate and the per-record dating in favour of an in-memory high-water mark that starts at zero, so the first session of every process exports an arbitrarily old stored weigh-in as today's live weight (the exact failure the QN 0xB4 fix already addressed in this repo), and the supported-scales row still describes the pre-refactor design and promises backlog collection that the code no longer does. The rest are docs-count and style nits plus the wrong base branch.

## Findings

### 1. Blocker: A stored record of any age is exported as today's weight on the first session of each process
`src/scales/salter.ts:297`

The only staleness guard is `lastReportedTs` (salter.ts:261), an in-memory high-water mark initialised to 0 on every process start, plus `isFromAnotherEpoch` (salter.ts:336-340) which rejects records only in the FUTURE direction. Nothing bounds how old an accepted record may be: 60 s ahead of the scale's clock is rejected, three days behind it is accepted. Commit 96a196d ("report only the newest reading, ignore the stored buffer") deliberately deleted `MAX_RECORD_AGE_SEC` and the per-record `timestamp:` alongside the history design, so the reading now leaves parseNotification undated (salter.ts:301). That also opts the reading out of the platform's own protections: src/runtime/processor.ts:180 sets `ExportContext.timestamp` only for timestamped readings, and the replay dedup at src/runtime/processor.ts:136-142 (weight vs `last_known_weight`) runs only for timestamped frames. The repo already carries this exact bug and its fix for the QN family: src/scales/qn-scale.ts:1004-1011 rejects a 0xB4 history record stamped before the session began, added because a stored record published 67.10 kg for a 75 kg user, and src/scales/qn-scale.ts:130 keeps `MAX_STORED_RECORD_AGE_SEC = 90` for the same reason.

**What goes wrong:** Verified by driving SalterAdapter through the real `waitForRawReading` seam with a simulated scale: fresh adapter, scale clock three days after every stored record, slots holding 87.6 / 89.7 / 88.8 kg. The session resolves with `{ weight: 89.7, impedance: 0 }`, `history: undefined`, no timestamp, so a three-day-old weigh-in is exported to Garmin/Strava/InfluxDB dated now. Real-world triggers: any systemd restart, container update or single-run `npm start` (every run is a fresh process, so the dedupe state is empty), where the connect happens while the scale is still awake from an already-exported weigh-in, or before the in-progress measurement has been written to a slot.

**Suggested fix:** Reinstate a maximum-age gate in parseNotification using the clock the adapter already tracks, mirroring qn-scale.ts: compute `clockNow = this.clockSec + (Date.now() - this.clockAt) / 1000` (as isFromAnotherEpoch already does) and return null when `clockNow - timestamp > MAX_RECORD_AGE_SEC` (QN uses 90 s), skipping the check only while `clockSec` is 0. Keep the high-water mark on top of it for the within-process repeat case. If instead the intent is to keep exporting old records, set `ScaleReading.timestamp` from the record so they route through the cache-replay path and are dated correctly.

### 2. Major: Supported-scales row promises backlog collection the adapter no longer does
`docs/guide/supported-scales.md:45`

The new row states that stored weigh-ins are "de-duplicated by their own timestamp and never cleared, so a backlog ... is collected in a single session and dated correctly". That describes the first commit's design. The branch's final commit (96a196d) replaced it: parseNotification emits no timestamp (salter.ts:301) and only records strictly newer than the high-water mark (salter.ts:297), and src/ble/shared.ts:511-516 hands each non-final complete reading to the HoldTimer, which keeps only the LAST one (src/ble/notification-processor.ts:72-84). So a session resolves with exactly one reading, undated, and every older record is not merely skipped but permanently suppressed because `lastReportedTs` has advanced past it, so it can never appear in a later session either. The author's own commit message says as much ("That guarantee is dropped deliberately"); only the docs were not updated.

**What goes wrong:** Harness run through waitForRawReading with three genuinely new records in one sweep (87.6, 89.7, 88.8 kg) resolves to a single RawReading of 89.7 kg with `history: undefined`. Concretely: person A weighs 87.6 kg, person B weighs 89.7 kg before the next sync runs; the docs say both are collected and dated, the code exports only B, undated, and A is lost for good.

**Suggested fix:** Rewrite the row to match the code: weight-only, the newest stored record wins, older stored records are ignored rather than replayed, and nothing is ever cleared on the scale. Move the protocol narrative into a Known Limitations entry the way R-MSC04, Eufy P2 and Koogeek S1 do (docs/guide/supported-scales.md:71-79) and keep the table cell to one or two sentences like every other row.

### 3. Minor: Adapter-count surfaces missed: logo easter egg and CONTRIBUTING
`docs/public/logo.svg:6`

The registry now holds 31 instances / 30 protocol adapters (verified by importing src/scales/index.ts), and the PR correctly bumps docs/guide/supported-scales.md:12, docs/alternatives.md:28 and docs/faq.md:65 to 30. Two surfaces the maintainer treats as part of the same set were missed. docs/public/logo.svg still reads 29.11; commit 87b6bbd ("docs: correct adapter and exporter counts") and cf85243 both bumped this easter egg as `adapters.exporters` in the same change as the prose, and 87b6bbd's message spells out the convention. CONTRIBUTING.md:52 still says "for all 30 adapters", which is the registry-instance count and is now 31.

**What goes wrong:** A reader comparing the site logo (29.11) with the supported-scales headline (30 protocol adapters) on the same deploy sees two different numbers, and the next person doing a count audit has to re-derive the convention.

**Suggested fix:** Change docs/public/logo.svg line 6 text to 30.11 and CONTRIBUTING.md:52 to "all 31 adapters".

### 4. Minor: Em dashes in new docs prose and double dashes in the PR/commit bodies
`docs/guide/supported-scales.md:45`

The project forbids the em dash and the doubled hyphen sequence in prose. The new table row introduces two em dashes into docs/guide/supported-scales.md, a file that currently contains zero. The PR body ("The six derived fields decode exactly", before "verified field by field") and commit 96a196d's body ("whichever slot the walk hit first") both use the doubled hyphen as a dash, and per CONTRIBUTING.md:286 the PR body becomes the squash commit body, so it lands in git history and in what release-please reads. The source comments in src/scales/salter.ts (16) and tests/scales/salter.test.ts (7) also use em dashes, but src/ already contains 110 of them across 47 files, so those are the maintainer's call rather than something this PR uniquely introduces.

**What goes wrong:** The row ships to blescalesync.dev with characters the project bans, and the squash commit body carries the banned dash into the release history.

**Suggested fix:** Rewrite the affected sentences without the dashes (the row is being rewritten for the backlog finding anyway) and edit the PR body before merge.

### 5. Minor: No upper plausibility bound on the decoded weight
`src/scales/salter.ts:288`

`weight = data.readUInt16LE(6) / 10` is accepted on `weight > 0 && Number.isFinite(weight)` alone, so the full u16 range up to 6553.5 kg passes. Every comparable adapter in the registry gates on a plausible band: src/scales/koogeek-s1.ts:32,152 (300 kg), src/scales/qn-scale.ts:246-247,1018 (5-300 kg), src/scales/eufy-p2.ts:104,121,347 (2-200 kg), src/scales/xiaomi-s800.ts:28. Only the low end needs care here, since the 7.4 kg fixture shows light loads are legitimate.

**What goes wrong:** Any 20-byte frame whose first byte is under 8 is treated as a record. On a firmware variant that answers some other opcode with a 20-byte reply, or on a garbled frame, bytes 6-7 of 0xFFFF with a non-zero stamp yield 6553.5 kg, which passes isComplete() and is exported.

**Suggested fix:** Add `const MAX_WEIGHT_KG = 300;` and reject `weight <= 0 || weight > MAX_WEIGHT_KG`, matching koogeek-s1.ts.

### 6. Minor: Bare 0xFFC1 characteristic claim widens the match surface above 13 adapters
`src/scales/salter.ts:179`

`charUuids: [CHR_FFC1]` makes matchesDescriptor return true on a characteristic hit alone (src/scales/match-descriptor.ts:83), with no service or name qualifier, at priority 125 which outranks Hoffen (20) through Exingtech Y1 (120). The post-discovery re-resolve paths feed the discovered characteristic list back through resolveAdapter (src/ble/handler-esphome-proxy/watcher.ts:174-176), so any device that exposes a SIG-based 0xFFC1 would be handed to Salter mid-session. No in-tree adapter or fixture collides today (I checked every serviceUuids/charUuids claim, and the TI OAD `f000ffc1-...` UUID in tests/ble/handler-mqtt-proxy.test.ts:2042 does not reduce to the 16-bit form so it does not hit), and the full suite passes, so this is hardening rather than a demonstrated break. Note also that the claim mostly cannot pay off: a unit that advertises neither the name nor 0xFFCC/0xCCFF would never reach GATT discovery as a Salter in the first place.

**What goes wrong:** A future or third-party scale that exposes a vendor 0xFFC1 characteristic and is matched pre-connect by a sub-125 adapter (Koogeek, Inlife, 1byone, Hutbit, MGB, Robi ...) is re-resolved to Salter after discovery, then fails the whole read because it does not speak the 0xFFCC protocol.

**Suggested fix:** Either drop `charUuids` (name plus both service forms already cover every observed advertisement) or mark the descriptor `custom: true` and require the FFCC/CCFF service alongside the characteristic in `matches()`.

### 7. Minor: PR targets main; contributions merge into dev
`CONTRIBUTING.md:281`

The PR's base branch is `main`. CONTRIBUTING.md:30 and :281 state that PRs and new features target `dev` and that contributors should branch from `dev`, not `main`; `main` is release-only and protected, and merging a feature straight into it bypasses the dev integration branch the release promotion relies on.

**What goes wrong:** Merged as-is, a feature commit lands directly on main outside the dev to main promotion flow, so dev and main diverge and the next promotion merge has to reconcile it.

**Suggested fix:** Retarget the PR to `dev` (the branch itself needs no changes).

## Worth calling out

- **Record layout is fully verifiable from the shipped fixtures** (src/scales/salter.ts:152) The documented layout (`<slot> 00 | u32 LE stamp | 7 x u16 LE, /10 except BMR`) reproduces exactly against the test fixtures. Decoding REC_857_APP by hand gives weight 0x0359 = 85.7 kg, fat 450 = 45.0 %, water 362 = 36.2 %, BMR 1078 kcal and BMI 857 = 85.7, which is precisely the 100 cm-profile anecdote in the class comment (85.7 / 1.00 m^2), and REC_897_SLOT2 gives BMI 310 = 89.7 / 1.70^2 for a different user slot. Every number in the prose is checkable from the committed bytes, which is rare and made this review much faster.

- **The destructive clear is proven absent, and the battery-change trap is handled and tested** (tests/scales/salter.test.ts:739) The `0a` clear opcode is documented but never emitted, and the test enumerates every write the adapter can produce (unlockCommand, unlockCommands and buildAck across all eight slots plus the clock reply) and asserts none starts with `0a`, which is the right shape of test for a rule about what must never happen. The clock-went-backwards handling in noteClock (salter.ts:313-322) catches the one failure mode that would make a high-water mark go permanently silent after a battery change, and the 'recovers after a battery change instead of going silent forever' test genuinely exercises it: deleting the reset makes it fail.

## Questions for you

- normalizesWeight is inferred, and shipping it as `true` is the right default here (scale.weight_unit is documented as display-only in docs/guide/configuration.md:167, and 32 of the 33 adapters declare the same), so this is not a blocker. Please still confirm it on hardware: set the scale to lb, take one weigh-in, and report the raw 20-byte record next to the kg value the scale displays. If bytes 6-7 still carry kg x 10 the inference is settled.

- A DEBUG log of one full real session (the `Salter RAW` lines from connect to `Reading complete`), so the sweep pacing, the number of writes per 3 s tick and the 5 s hold resolving on the newest slot can be seen on real hardware rather than in a simulator.

- On the hardware: is the record for the weigh-in in progress written to a slot BEFORE or AFTER the scale becomes connectable? That decides how often the first session of a process picks up the PREVIOUS weigh-in instead of the current one, which is the blocker above.

- Do you want the freshness gate reinstated (newest-wins plus a max age against the scale's own clock), or the timestamped history path restored so older stored records export dated? Either resolves the blocker; the docs row should then be rewritten to match whichever you pick.

KristianP26:
Checking in on this one, and flagging two things that changed under it.

**The blocker still stands.** The only staleness guard is `lastReportedTs`, an in-memory high-water mark that starts at zero on every process start, plus a check that rejects records only in the future direction. Nothing bounds how old an accepted record may be, so the first session of each process exports whatever stored weigh-in the scale still holds as today's live weight. Since the reading also leaves `parseNotification` undated, it opts out of the platform's own protections: the export context only carries a timestamp for timestamped readings, and the replay dedup against `last_known_weight` runs only for timestamped frames. This repo already carries this exact bug and its fix for the QN family, where a stored record published 67.10 kg for a 75 kg user.

The decode itself I have no concerns about. I checked every field claim arithmetically against your fixtures and they all hold, and the app log you attached is a good capture.

**Two things moved on `dev` since the review:**

1. The adapter count. `#353` added an adapter and I added another, so the registry is at 32 entries. Your doc hunks still bump the counts to the values that were correct when you opened this. They will need to go one further: 31 protocol adapters in `docs/guide/supported-scales.md` and `docs/alternatives.md`, 32 in the `docs/faq.md` JSON-LD, `31.11` in `docs/public/logo.svg`, 31 in `CONTRIBUTING.md`. Git will merge your version cleanly and leave the numbers wrong, so this is worth doing by hand.
2. `tests/scales/registry-collision.test.ts` gained two fixtures around the line your hunk lands on, so expect a small conflict there.

Also two housekeeping items before this can merge: the supported-scales row still describes the pre-refactor design and promises backlog collection the code no longer does, and the PR body contains a prose double dash plus the unfilled template comments and an unticked checkbox. The body becomes the squash commit message verbatim, so it is worth a tidy.

No rush on my side. If you would rather not carry the staleness work, say so and I will take it as a follow-up rather than leaving the PR sitting.

KristianP26:
I took the staleness work rather than leaving this parked, since you had not had a chance to come back to it and everything else here was ready.

## What I changed

One commit on top of yours: `fix(salter): bound how old a stored record may be before it is reported`.

A record is now rejected unless it sits within five minutes of the scale's own clock, and rejected outright until that clock has been read, because without it there is nothing to measure age against. Both gates are mutation-tested: removing the age bound fails two tests, removing the no-clock gate fails one.

Five minutes is deliberately loose rather than tight. The sweep restarts every few seconds, but a continuous-mode run reconnects on its own cadence, and someone who steps off before the host connects should still get their reading.

## Where I was wrong in the first review

I said the reading should carry a `timestamp` so it would pick up the platform's replay protections. **You were right to remove it and I was wrong to ask for it.** I checked before changing anything: a dated reading is pushed into the cache-replay buffer in `waitForRawReading`, which returns early and only drains on disconnect, while this adapter holds the link open and polls until the session times out, and a timeout rejects. A dated reading would be buffered and never delivered. Your test comment says exactly that, and it is correct. The reading stays undated and the age bound stands in for those protections instead. I have written that reasoning into the code so nobody undoes it later, including me.

## One test I restamped

`emits an ascending sweep in order` fed four captures spanning 68 minutes and expected all four to emit. Under the age bound the older ones are history, so as written it asserted the behaviour being fixed. The records are restamped inside the window; the ordering it actually tests is unchanged.

## Housekeeping I did rather than sending back

- Rebased onto `dev`, which had moved a long way.
- The adapter count needed to go two further than your hunks: two adapters landed today. Registry is 33 entries, so 32 protocol adapters, `32.11` in the logo, 33 in the FAQ JSON-LD.
- A conflict resolution of mine briefly left markers inside the Known Limitations table; caught and removed before this push.
- The supported-scales row still described the pre-refactor design and promised backlog collection the code no longer does. Rewritten to match what it now does.

Full suite on the rebased head: 134 files, 2321 tests, tsc, eslint and prettier clean.

The decode was never in question. I checked every field arithmetically against your fixtures in the first pass and they all hold. Thank you for the adapter, the app log and the DIS dump.

PR #347, opened 2026-08-23
fix(notifications): format weight, muscle and bone in the configured weight unit

### LINKED ISSUE #342: ntfy and telegram notifications hardcode kg and ignore scale.weight_unit
### Description

I run with `scale.weight_unit: lbs`. The console prints `Measurement: 199.08 lbs`, the Garmin upload is right, and the ntfy notification for the same weigh-in says `⚖️ 90.30 kg | ... Muscle 62.4 kg | ... Bone 3.1 kg`. Telegram sends the same message. Both exporters build the text in their own `formatMessage` with a literal `kg`, while the console goes through `fmtWeight()` in `src/runtime/format.ts`, which already knows the unit. The setting never reaches the exporters because `ExportContext` doesn't carry it.

PR follows this issue.

### Steps to Reproduce

1. `scale.weight_unit: lbs` and an ntfy or telegram exporter in `global_exporters`
2. Weigh in
3. Compare the `Measurement:` log line with the notification body

### Scale Brand / Model

Renpho ES-CS20M, but it's the exporters, any scale does it

### Platform

Docker (Linux, mqtt-proxy via ESP32-S3)

### Node.js Version

image default (v1.24.0 container)

### App Version

1.24.0

### Debug Logs

```
[Sync] Measurement: 199.08 lbs / 513 Ohm
```

Notification body for the same reading:

```
⚖️ 90.30 kg | BMI 25.8
🏋️ Body Fat 18.5% | Muscle 62.4 kg
💧 Water 55.2% | 🦴 Bone 3.1 kg
```

alasano:
Closes #342.

The ntfy and telegram exporters each had their own copy of the message formatter with `kg` hardcoded, and the console was already formatting the same three fields through `fmtWeight()`, which handles `scale.weight_unit`. The processor now puts the unit in `ExportContext.weightUnit`, and the two formatters, which have been identical apart from telegram's title line since telegram was added, became one `formatNotification()` in `src/exporters/notification-message.ts` that calls `fmtWeight()`. The notification now says what the log line says.

Muscle and bone go from one decimal to two as a side effect, because `fmtWeight()` prints two and the console already did.

I added an lbs case for each exporter and the processor test checks the context carries the unit. One sentence in each exporter's doc section. Full suite passing except the two pre-existing macOS-only failures (BF720 date of birth, config watcher), tsc/eslint/prettier clean.

### DIFF
--- a/docs/exporters.md
+++ b/docs/exporters.md
@@ -184,6 +184,8 @@ global_exporters:
     priority: 4
 ```
 
+Weight, muscle and bone follow `scale.weight_unit`.
+
 ## Telegram {#telegram}
 
 Send a measurement notification to a Telegram chat via a bot. Create a bot with [@BotFather](https://t.me/BotFather) to get a bot token, then start a chat with your bot (or add it to a group/channel) so it can message you.
@@ -204,7 +206,7 @@ global_exporters:
     silent: false
 ```
 
-The message is sent as plain text. In multi-user setups the user's name is prepended as `[Name]`. Historical readings replayed from a scale's offline cache are skipped (a notification for an old measurement is not meaningful).
+The message is sent as plain text. Weight, muscle and bone follow `scale.weight_unit`. In multi-user setups the user's name is prepended as `[Name]`. Historical readings replayed from a scale's offline cache are skipped (a notification for an old measurement is not meaningful).
 
 ::: tip Finding your chat ID
 Message your bot once, then open `https://api.telegram.org/bot<token>/getUpdates` in a browser — the `chat.id` field holds your chat ID. For groups, add the bot to the group first.
--- a/src/exporters/notification-message.ts
+++ b/src/exporters/notification-message.ts
@@ -1,0 +1,20 @@
+import type { BodyComposition } from '../interfaces/scale-adapter.js';
+import type { ExportContext } from '../interfaces/exporter.js';
+import { fmtWeight } from '../runtime/format.js';
+
+/** Message body shared by the push notification exporters (ntfy, telegram). */
+export function formatNotification(data: BodyComposition, context?: ExportContext): string {
+  const unit = context?.weightUnit ?? 'kg';
+  const prefix = context?.userName ? `[${context.userName}] ` : '';
+  const lines = [
+    `${prefix}⚖️ ${fmtWeight(data.weight, unit)} | BMI ${data.bmi.toFixed(1)}`,
+    `🏋️ Body Fat ${data.bodyFatPercent.toFixed(1)}% | Muscle ${fmtWeight(data.muscleMass, unit)}`,
+    `💧 Water ${data.waterPercent.toFixed(1)}% | 🦴 Bone ${fmtWeight(data.boneMass, unit)}`,
+    `🫀 Visceral Fat ${data.visceralFat} | BMR ${data.bmr} kcal`,
+    `📅 Metabolic Age ${data.metabolicAge} yr | Physique ${data.physiqueRating}`,
+  ];
+  if (context?.driftWarning) {
+    lines.push(`⚠️ ${context.driftWarning}`);
+  }
+  return lines.join('\n');
+}
--- a/src/exporters/ntfy.ts
+++ b/src/exporters/ntfy.ts
@@ -3,6 +3,7 @@ import type { BodyComposition } from '../interfaces/scale-adapter.js';
 import type { Exporter, ExportContext, ExportResult } from '../interfaces/exporter.js';
 import type { ExporterSchema } from '../interfaces/exporter-schema.js';
 import type { NtfyConfig } from './config.js';
+import { formatNotification } from './notification-message.js';
 import { withRetry, httpError } from '../utils/retry.js';
 import { errMsg } from '../utils/error.js';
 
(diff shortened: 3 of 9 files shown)


KristianP26 (approved):
The PR collapses two near-identical message formatters in src/exporters/ntfy.ts and src/exporters/telegram.ts into one src/exporters/notification-message.ts that routes weight, muscleMass and boneMass through the existing fmtWeight(), and threads scale.weight_unit to the exporters via a new optional ExportContext.weightUnit set in src/runtime/processor.ts:179. The plumbing is sound: ctx.weightUnit is a required field on AppContext (src/runtime/context.ts:18, populated at creation and re-set on hot reload), src/runtime/processor.ts:175 is the only ExportContext construction site in src/, and src/orchestrator.ts:88 is the only e.export() call site, so no path can reach these exporters with the unit missing. Historical replay never reaches ntfy or telegram at all because neither declares supportsBackdate and src/orchestrator.ts:59 filters them out first. The muscle and bone precision change from one to two decimals is intentional and makes the notification byte-identical to the console line, which has printed those two through fmtWeight since KG_METRICS was added. Verified on the PR head: tsc, eslint, prettier clean, 2217 tests pass, and the new lbs tests genuinely fail when the pre-change formatters are restored.

## Findings

### 1. Minor: Drift warning line still says kg, so an lbs notification mixes units
`src/exporters/notification-message.ts:17`

The five metric lines now honour context.weightUnit, but the appended warning is passed through verbatim from detectWeightDrift(), which hardcodes kg at src/config/user-matching.ts:36, 40 and 44 and prints the stored range bounds (always kg, converted at src/wizard/steps/users.ts:94). The new not.toContain('kg') test does not catch this because neither lbs test supplies a driftWarning. This is defensible as out of scope, since the console log at src/runtime/processor.ts (log.warn(`${prefix} ${drift}`)) has exactly the same wording, so the PR's own claim that the notification matches the log still holds. Worth a follow-up rather than a change here.

**What goes wrong:** Multi-user config with scale.weight_unit: lbs, Dad's weight_range [75, 95] kg, a 94 kg weigh-in lands in the upper 10%. The telegram message reads: '⚖️ 207.23 lbs | BMI 28.0' on line 2 and '⚠️ Weight 94 kg is near the upper boundary of Dad's range [75-95]' on the last line, so one message carries both units.

**Suggested fix:** Leave as is for this PR and open a follow-up that gives detectWeightDrift a WeightUnit parameter and formats both the measured weight and the two range bounds through fmtWeight(), so the console warning and the notification warning move together.

### 2. Minor: Strava success log is the one remaining human-readable hardcoded kg
`src/exporters/strava.ts:77`

Scanning every exporter for hardcoded units: file.ts, influxdb.ts, webhook.ts, garmin.ts, runalyze.ts, wger.ts and intervals.ts all emit machine payloads in kg and should stay that way; mqtt.ts:28/39/40 set unit_of_measurement metadata that Home Assistant converts itself. The only other human-readable string is this Strava log line, which prints toFixed(2) with a literal kg. It is arguably correct as written, because the value it reports is what was PUT to the Strava API (body: JSON.stringify({ weight: data.weight }), line 69), and Strava's athlete weight field is kg. Calling it out only because the export signature deliberately ignores context (_context) and the new weightUnit field makes a one-line change possible if you want the console fully consistent.

**What goes wrong:** With scale.weight_unit: lbs, the console prints '[Sync] Measurement: 199.08 lbs' followed by '[Strava] Strava weight updated to 90.30 kg.' for the same weigh-in, which is the same console/notification mismatch #342 reported, on a different surface.

**Suggested fix:** Optional: rename _context to context and log `Strava weight updated to ${fmtWeight(data.weight, context?.weightUnit ?? 'kg')}.`, or leave it and reword to make clear it is reporting the API payload, e.g. 'Strava athlete weight set to 90.30 kg'.

### 3. Minor: First import from src/exporters into src/runtime
`src/exporters/notification-message.ts:3`

src/runtime/exporters.ts:2 already imports from src/exporters/registry.js, so this line makes the directory dependency bidirectional. I checked for an actual cycle and there is none: format.ts is a leaf whose only import is a type-only `import type { WeightUnit } from '../config/schema.js'`, which is erased, and madge --circular on the PR head reports only the pre-existing interfaces/scale-adapter.ts > scales/match-descriptor.ts pair. So this is taste, not a defect, and reusing fmtWeight rather than duplicating the conversion is clearly the right call.

**What goes wrong:** No runtime failure. The risk is future-facing: if src/runtime/format.ts ever grows a value import back into src/exporters or src/runtime/context.js, the exporters barrel starts pulling runtime state through a formatting helper.

**Suggested fix:** Optionally move fmtWeight and KG_TO_LBS to src/utils/format.ts (alongside error.ts and retry.ts) and re-point the four existing call sites in src/runtime/processor.ts and src/runtime/poll-source.ts, so the shared formatter lives in the layer both sides already depend on.

## Worth calling out

- **The lbs tests are genuinely load-bearing, not decorative** (tests/exporters/telegram.test.ts:88) I restored the pre-change ntfy.ts and telegram.ts under the new test files and re-ran: 4 assertions fail, with 'expected Scale Measurement\n⚖️ 80.00 kg | BMI ... to contain ⚖️ 176.37 lbs'. The not.toContain('kg') assertion is also not vacuous in either direction: the message still contains 'BMR 1750 kcal', which does not match the substring 'kg', and the same assertion fails on the unfixed code because the body carries '80.00 kg'. The arithmetic checks out against fmtWeight's KG_TO_LBS = 2.20462 (80 -> 176.37, 62.4 -> 137.57, 3.1 -> 6.83). The processor test is equally tight: the single-user case uses toEqual, so the exact ExportContext shape including weightUnit: 'kg' is pinned, and the multi-user case overrides makeCtx to 'lbs' so a hardcoded default would not pass.

PR #348, opened 2026-08-23
fix(beurer): provision the date of birth in UTC so the day does not shift west of Greenwich

### LINKED ISSUE #344: BF720: date of birth is provisioned one day early on hosts west of UTC
### Description

`src/scales/beurer-bf720.ts` builds the date-of-birth characteristic with `new Date(p.birthDate)` and then `getFullYear()` / `getMonth()` / `getDate()`. `birth_date` is validated as `YYYY-MM-DD`, which `Date` parses as UTC midnight, and the local-time getters read that back as the previous day anywhere west of Greenwich. Someone born 1990-06-15 on a host in America/Montreal gets `1990-06-14` written to the scale, so the age the scale uses for body composition is off by one on their birthday.

I don't own a BF720. I found this because `tests/scales/beurer-bf720.test.ts` > "provisions the real date of birth when the profile carries one" fails on my Mac in UTC-4 and passes on CI in UTC. The time-of-day write in the same file uses the local getters on `new Date()` and that one is right as is.

PR follows this issue.

### Steps to Reproduce

1. `TZ=America/Montreal npx vitest run tests/scales/beurer-bf720.test.ts`
2. `expected 'c607060e' to be 'c607060f'` (day byte 14 instead of 15)
3. `TZ=UTC` passes

### Scale Brand / Model

Beurer BF720 (adapter code, not tested on hardware)

### Platform

macOS 26 (dev machine), but it would hit any host in a negative UTC offset

### Node.js Version

v26

### App Version

1.24.0

### Debug Logs

```
FAIL  tests/scales/beurer-bf720.test.ts > BeurerBf720Adapter > session hygiene and provisioning bounds (#229) > provisions the real date of birth when the profile carries one
AssertionError: expected 'c607060e' to be 'c607060f' // Object.is equality
```

alasano:
Closes #344.

`birth_date` is validated as `YYYY-MM-DD`, which `Date` parses as UTC midnight, and the provisioning code read it back with the local-time getters, so any host west of UTC wrote the previous day. It now reads the parts back with `getUTCFullYear` / `getUTCMonth` / `getUTCDate`. The time-of-day write in the same adapter stays on local time, that one is the wall clock.

The existing test passes in America/Montreal and in UTC+14 now, it was already asserting the right bytes. tsc/eslint/prettier clean.

### DIFF
--- a/src/scales/beurer-bf720.ts
+++ b/src/scales/beurer-bf720.ts
@@ -64,10 +64,12 @@ const PROFILE_FIELDS: ProfileField[] = [
       return year >= 1900 && year <= new Date().getFullYear();
     },
     fromProfile: (p) => {
+      // YYYY-MM-DD parses as UTC midnight, so read it back in UTC: the local
+      // getters shift the day west of Greenwich.
       const born = p.birthDate ? new Date(p.birthDate) : null;
       if (born && !Number.isNaN(born.getTime())) {
-        const y = born.getFullYear();
-        return [y & 0xff, (y >> 8) & 0xff, born.getMonth() + 1, born.getDate()];
+        const y = born.getUTCFullYear();
+        return [y & 0xff, (y >> 8) & 0xff, born.getUTCMonth() + 1, born.getUTCDate()];
       }
       // No birth date in the profile: anchor to 1 January of the derived year.
       const year = new Date().getFullYear() - Math.max(1, Math.round(p.age));
--- a/tests/scales/beurer-bf720.test.ts
+++ b/tests/scales/beurer-bf720.test.ts
@@ -918,6 +918,31 @@ describe('BeurerBf720Adapter', () => {
       expect(Buffer.from(dob![1] as number[]).toString('hex')).toBe('c607060f');
     });
 
+    // Local and UTC getters agree at or east of Greenwich, so CI (UTC) cannot
+    // see #344. Pin the zone or this branch is effectively untested.
+    it('provisions the calendar date of birth on a host west of Greenwich', async () => {
+      const savedTz = process.env.TZ;
+      process.env.TZ = 'America/Montreal';
+      try {
+        const a = makeAdapter();
+        const ctx = ctxWith(
+          { [DOB]: Buffer.from('00000000', 'hex') },
+          { provision: true, profile: defaultProfile({ birthDate: '1990-06-15' }) },
+        );
+        await a.onConnected(ctx);
+        const write = ctx.write as ReturnType<typeof vi.fn>;
+        write.mockClear();
+        a.parseCharNotification(UCP, Buffer.from('200201', 'hex'));
+        await settle();
+        const dob = write.mock.calls.find((c) => c[0] === DOB);
+        // 1990-06-15, not the 14th the local getters read back at UTC-4.
+        expect(Buffer.from(dob![1] as number[]).toString('hex')).toBe('c607060f');
+      } finally {
+        if (savedTz === undefined) delete process.env.TZ;
+        else process.env.TZ = savedTz;
+      }
+    });
+
     it('does not claim the scale is empty once the consent was accepted', () => {
       const warn = vi.spyOn(bleLog, 'warn').mockImplementation(() => {});
       const a = makeAdapter();


KristianP26 (changes requested):
The fix is correct and minimal: `birth_date` is regex-pinned to `YYYY-MM-DD` (src/config/schema.ts:185, and src/validate-env.ts:118 for the .env path), `resolveUserProfile` is the only producer of `UserProfile.birthDate`, so `new Date(p.birthDate)` is always a date-only string parsed at UTC midnight and the UTC getters are the right way to read it back. I checked the two things that could have made the UTC/local mix inconsistent and neither bites: `isSet`'s `new Date().getFullYear()` is an upper bound on "now" rather than on the profile date, and the no-birthdate fallback derives its year from local "now" exactly like `computeAge` in src/config/resolve.ts:20. There is no read-back path that parses a date of birth into a `Date`, and `buildCurrentTime` / `parseTimestamp` correctly stay on wall clock. The one gap is coverage: CI runs ubuntu-latest with no `TZ`, so the existing test passes on the unpatched code and the regression can silently come back.

## Findings

### 1. Major: The regression is not detectable in CI, so nothing pins the fix
`tests/scales/beurer-bf720.test.ts:905`

The only test that reaches this branch is 'provisions the real date of birth when the profile carries one', which asserts c607060f for birthDate '1990-06-15'. Local getters and UTC getters agree on every host at or east of Greenwich, and .github/workflows/ci.yml runs ubuntu-latest with no TZ set, so the runner is UTC. I confirmed this empirically: with the PR reverted on disk, `npx vitest run tests/scales/beurer-bf720.test.ts -t 'provisions the real date of birth'` passes (1 passed, 64 skipped). Issue #344 itself was only found because the reporter's dev machine happens to sit at UTC-4. The PR body says the test 'passes in America/Montreal and in UTC+14 now', which was verified by hand and is not encoded anywhere the suite will re-check.

**What goes wrong:** A later cleanup pass normalises the four date getters in this file back to the local forms (they sit five lines from `buildCurrentTime`, which legitimately uses `getFullYear`/`getMonth`/`getDate`), or the field moves into a shared profile helper. The full suite stays green on CI, prettier/eslint/tsc stay clean, and every BF720/BF788/BF950 user in the Americas is silently provisioned with 1990-06-14 instead of 1990-06-15 again, shifting the age the scale uses for its own body-composition maths by one year around their birthday.

**Suggested fix:** Add a sibling test in the same `describe('session hygiene and provisioning bounds (#229)')` block that pins the zone. Runtime assignment of `process.env.TZ` does reconfigure Node's local zone (verified on Node 24 here), so:

```ts
// Local and UTC getters agree at or east of Greenwich, so CI (UTC) cannot
// see #344. Pin the zone or this branch is effectively untested.
it('provisions the calendar date of birth on a host west of Greenwich', async () => {
  const savedTz = process.env.TZ;
  process.env.TZ = 'America/Montreal';
  try {
    const a = makeAdapter();
    const ctx = ctxWith(
      { [DOB]: Buffer.from('00000000', 'hex') },
      { provision: true, profile: defaultProfile({ birthDate: '1990-06-15' }) },
    );
    await a.onConnected(ctx);
    const write = ctx.write as ReturnType<typeof vi.fn>;
    write.mockClear();
    a.parseCharNotification(UCP, Buffer.from('200201', 'hex'));
    await settle();
    const dob = write.mock.calls.find((c) => c[0] === DOB);
    // 1990-06-15, not the 14th the local getters read back at UTC-4.
    expect(Buffer.from(dob![1] as number[]).toString('hex')).toBe('c607060f');
  } finally {
    if (savedTz === undefined) delete process.env.TZ;
    else process.env.TZ = savedTz;
  }
});
```

It asserts the day byte is 0x0f rather than the 0x0e the pre-fix code produces, and it fails on the current main.

## Worth calling out

- **The UTC read also realigns the provisioned date with the age the app computes** (src/scales/beurer-bf720.ts:71) `computeAge` in src/config/resolve.ts:19-26 splits the birth date string into its literal components and never parses it as a `Date`, so it has always used the calendar day the user typed. Reading the same string back with the UTC getters returns exactly those literal components, so the age the app derives and the date of birth the scale stores now agree on every host. Before this change they disagreed by a day west of UTC. The scope of the change is right too: `buildCurrentTime` (line 533) and `parseTimestamp` (line 641) are both wall clock and were correctly left alone, and this was the only place in src/ that parsed an ISO date string and then read it with local getters (the other local getters all operate on `new Date()` or on locally constructed dates in src/validate-env.ts and src/wizard/steps/users.ts). No docs change is needed: docs/guide/configuration.md already documents birth_date as an ISO YYYY-MM-DD date, no config surface moved, and no stated count changed.

## Questions for you

- Please add the timezone-pinned regression test described above so CI can see the bug. Everything else in the change is verified correct and I would merge it as soon as that test is in.

- Confirm whether you ran the full suite (not just tests/scales/beurer-bf720.test.ts) under TZ=America/Montreal, so we know the pinned test is the only timezone-sensitive assertion being introduced.

alasano:
Added the timezone-pinned test, pretty much what you wrote out. It sets `TZ=America/Montreal`, provisions `1990-06-15`, and asserts the day byte is `0x0f`. It fails on the current code with `0x0e` and passes with the UTC read.

I ran the full suite under `TZ=America/Montreal`, not just the BF720 file. Everything passes except the config-watcher test about sibling-file edits, and that one is the macOS FSEvents thing from the other issue, it fails on my Mac with or without this change. So the new test is the only timezone-sensitive assertion this PR adds.

KristianP26 (approved):
Verified the new test does what it claims: with the UTC getters reverted on disk it fails with `expected 'c607060e' to be 'c607060f'`, so the regression is now pinned on a UTC runner. Merging.

PR #349, opened 2026-08-23
fix(config): ignore watcher events that leave the config content unchanged

### LINKED ISSUE #345: config watcher fires on the file's own creation on macOS (FSEvents replay), test fails locally
### Description

Minor. `tests/config/watch.test.ts` > "ignores edits to other files in the same directory" fails on macOS. The filename filter in `src/config/watch.ts` is fine, the cause is that `fs.watch` on macOS (FSEvents) replays directory events from just before the watch was created, so the `config.yaml` written in `beforeEach` reaches the watcher as an edit and `onChange` fires once. A small probe script shows the replayed `rename config.yaml` arriving about 5 ms after the watcher starts. Linux inotify doesn't replay, so CI is green.

The same replay happens at runtime on macOS, one redundant `Config reloaded` right after startup. Harmless, same cause.

The fix I went with is to drop watcher events that leave the file content unchanged, one small read per event and events are rare. Content rather than mtime so a filesystem with one-second timestamps doesn't drop a real edit made in the same second as startup. The test passes without any sleep in the setup. PR follows this issue.

### Steps to Reproduce

1. On macOS, `npx vitest run tests/config/watch.test.ts`
2. `expected "vi.fn()" to not be called at all, but actually been called 1 times`

### Scale Brand / Model

n/a

### Platform

macOS 26 (dev machine)

### Node.js Version

v26

### App Version

1.24.0

### Debug Logs

```
FAIL  tests/config/watch.test.ts > startConfigWatcher > ignores edits to other files in the same directory
AssertionError: expected "vi.fn()" to not be called at all, but actually been called 1 times
```

alasano:
Closes #345. This is a small PR that makes the suite pass on macOS and removes one redundant reload at startup there. It's not required, but it makes all tests pass on Mac.

FSEvents replays directory history from just before a watch starts, so the config file's own creation reached the watcher as an edit. The watcher now reads the file on each event for its basename and drops the event when the content is what it last saw. A missing file (mid atomic rename, deleted) counts as changed, so tmp+rename and delete-and-recreate still fire. Content rather than mtime, because HFS+ and FAT store whole seconds and an edit in the same second as startup would be lost.

One new test, a rewrite with identical content doesn't fire, and the sibling-file test passes on macOS without touching it. Full suite passing except the pre-existing BF720 date-of-birth failure, tsc/eslint/prettier clean.

### DIFF
--- a/src/config/watch.ts
+++ b/src/config/watch.ts
@@ -1,4 +1,4 @@
-import { watch, type FSWatcher } from 'node:fs';
+import { watch, readFileSync, type FSWatcher } from 'node:fs';
 import { dirname, basename } from 'node:path';
 import { createLogger } from '../logger.js';
 import { errMsg } from '../utils/error.js';
@@ -19,6 +19,10 @@ export interface ConfigWatcherHandle {
  * inode, plus editor save patterns (vim `:w`, VS Code) that trigger 2+ events
  * within ~50 ms.
  *
+ * Events that leave the file's content unchanged are ignored: FSEvents on macOS
+ * replays directory history from just before the watch started, so the file's
+ * own creation would otherwise arrive as an edit.
+ *
  * Self-writes from updateLastKnownWeight() are suppressed via the suppress
  * window in write.ts, so this never re-fires for our own bumps. Errors from
  * fs.watch (e.g. parent directory unmounted) are logged and the watcher
@@ -29,6 +33,7 @@ export function startConfigWatcher(configPath: string, onChange: () => void): Co
   const base = basename(configPath);
   let debounceTimer: ReturnType<typeof setTimeout> | null = null;
   let closed = false;
+  let lastContent = readContent(configPath);
 
   const fire = () => {
     debounceTimer = null;
@@ -47,6 +52,9 @@ export function startConfigWatcher(configPath: string, onChange: () => void): Co
       // cannot tell whether the change is for our config file, so ignore.
       if (!filename) return;
       if (filename !== base) return;
+      const content = readContent(configPath);
+      if (content === lastContent) return;
+      lastContent = content;
       if (debounceTimer) clearTimeout(debounceTimer);
       debounceTimer = setTimeout(fire, DEBOUNCE_MS);
     });
@@ -78,4 +86,13 @@ export function startConfigWatcher(configPath: string, onChange: () => void): Co
       }
     },
   };
+}
+
+/** Current file content, or null while it is absent (mid atomic rename, deleted). */
+function readContent(path: string): string | null {
+  try {
+    return readFileSync(path, 'utf8');
+  } catch {
+    return null;
+  }
 }
--- a/tests/config/watch.test.ts
+++ b/tests/config/watch.test.ts
@@ -75,6 +75,19 @@ describe('startConfigWatcher', () => {
     }
   });
 
+  it('ignores events that leave the content unchanged', async () => {
+    const onChange = vi.fn();
+    const handle = startConfigWatcher(configPath, onChange);
+
+    try {
+      writeFileSync(configPath, 'version: 1\n');
+      await new Promise((r) => setTimeout(r, SETTLE_MS));
+      expect(onChange).not.toHaveBeenCalled();
+    } finally {
+      handle.close();
+    }
+  });
+
   it('survives atomic rename (tmp+rename pattern)', async () => {
     const onChange = vi.fn();
     const handle = startConfigWatcher(configPath, onChange);


KristianP26 (approved):
The watcher now snapshots the file content at startup and drops any event for the config basename whose re-read content is byte-identical, which correctly kills the macOS FSEvents replay of the file's own creation (#345) without touching the filename filter or the debounce. I could not construct a lost-edit case: every write syscall produces its own event and the callback's read can only observe the file at or after that write, so even an adversarially torn write whose intermediate state is byte-identical to the previous content still fires (probed on Windows: truncate, rewrite the old bytes, pause 300 ms, append, reload fired). The self-write concern is also refuted: the compare sits in the watch callback ahead of fire()'s isReloadSuppressed() check, so a suppressed self-write still refreshes the snapshot (probed: after a suppressed 80 to 82 kg write, reverting the file to the startup content still fired). The new test genuinely fails against unpatched dev; tsc, eslint and prettier are clean and tests/config plus tests/runtime pass (330 tests). Only nits below.

## Findings

### 1. Minor: Nothing pins the invariant that a suppressed self-write still refreshes the snapshot
`tests/config/watch.test.ts:113`

The fix stays correct under self-writes only because the compare lives in the watch callback (src/config/watch.ts:55-57) and not in fire(), which is where the isReloadSuppressed() check already sits. That placement is load-bearing but invisible: the existing suppress-window test at line 113 writes content that is never read back, so it passes either way. A test that reverts the file to the exact startup content after a suppressed self-write would lock it in, and it is fast if the window is shorter than SETTLE_MS.

**What goes wrong:** Someone later moves `const content = readContent(configPath); if (content === lastContent) return;` down into fire(), below the suppress check, as a natural tidy-up. updateLastKnownWeight then bumps last_known_weight from 80 to 82, the fire is suppressed, and lastContent stays at the startup content. The user restores a config.yaml backup whose bytes match that startup content: the event reads content identical to the stale snapshot, the reload never fires, and the whole edit is silently dropped. The suite stays green.

**Suggested fix:** Add after the existing suppress-window test:

  it('still fires for an edit that reverts to the pre-self-write content', async () => {
    const onChange = vi.fn();
    const handle = startConfigWatcher(configPath, onChange);
    try {
      setSuppressReloadWindow(600);
      writeFileSync(configPath, 'version: 1\nselfwrite: true\n');
      await new Promise((r) => setTimeout(r, SETTLE_MS));
      expect(onChange).not.toHaveBeenCalled();
      writeFileSync(configPath, 'version: 1\n'); // back to the startup bytes
      await new Promise((r) => setTimeout(r, SETTLE_MS));
      expect(onChange).toHaveBeenCalledTimes(1);
    } finally {
      handle.close();
    }
  });

### 2. Minor: readContent swallows every error, so a persistent read failure silences the watcher without a log line
`src/config/watch.ts:95`

The catch maps ENOENT (the intended case, documented on line 91) and every other error to the same null. A transient failure is harmless because the next event reads real content again, but a persistent one latches null into lastContent, after which every further event hits `content === lastContent` and returns early. The doc comment only mentions the absent-file case, so the broader swallow is easy to miss.

**What goes wrong:** config.yaml is chown'd to root with mode 600 while the service runs as a normal user (Pi systemd unit). First event: readFileSync throws EACCES, readContent returns null, the reload fires once and logs the read error. Every later edit reads null again, null === lastContent, and the callback returns before arming the debounce: the watcher is permanently silent with nothing in the log explaining why edits stopped applying. Before this PR each edit produced a fresh, visibly failing reload attempt.

**Suggested fix:** Log the swallowed error so the silence is diagnosable, and widen the doc comment:

/** Current file content, or null when it cannot be read (absent mid atomic rename, deleted, unreadable). */
function readContent(path: string): string | null {
  try {
    return readFileSync(path, 'utf8');
  } catch (err) {
    log.debug(`Config read during watch event failed: ${errMsg(err)}`);
    return null;
  }
}

### 3. Minor: Live Config Reload docs still imply every save is detected
`docs/multi-user.md:100`

Line 100 says edits to config.yaml are detected automatically. That is now qualified: a save that leaves the file byte-identical is deliberately ignored, which is the whole point of the PR but is not stated anywhere user-facing. One sentence keeps the page accurate. No count or table in docs/guide/configuration.md, docs/exporters.md or docs/guide/supported-scales.md is affected by this PR, and watch_config on line 230 of configuration.md still reads correctly.

**What goes wrong:** A user re-saves config.yaml unchanged, or runs touch config.yaml, expecting the documented auto-detect to force a reload after changing something outside the file. Nothing happens and nothing is logged, and the page gives no hint why. SIGHUP still works, so this is a wording gap rather than a functional one.

**Suggested fix:** Append to the paragraph on line 100: 'A save that leaves the file byte-identical is ignored; use SIGHUP to force a reload.'

## Worth calling out

- **Correct placement of the compare, and content rather than mtime** (src/config/watch.ts:55) Three things are right here. The compare is in the watch callback, above fire(), so a self-write suppressed by write.ts still refreshes lastContent, which is exactly the trap this design could have fallen into. Content rather than mtime avoids the whole-second timestamp problem on HFS+ and FAT that the PR body calls out. And a missing file compares as changed, so the unlink window of atomicWrite and delete-and-recreate with new content still fire, which the existing atomic-rename test confirms. The sync read is also gated behind the basename filter, so sibling churn costs nothing, and a 2 MB config read measured as noise against the 500 ms debounce.

- **The new test really does pin the changed branch** (tests/config/watch.test.ts:78) I checked the obvious failure mode for this kind of test, that the debounce or the suppress window already swallowed the event, and it does not apply: _resetSuppressWindow() runs in beforeEach and the test sets no window, so fire() reaches onChange.

PR #350, opened 2026-08-23
feat(notifications): report the other exporters' results with report_exports

### LINKED ISSUE #343: ntfy/telegram: option to wait for the other exporters and include their results in the notification
### Description

The ntfy and telegram exporters run at the same time as the data exporters, so the notification goes out the moment the scale is read and says nothing about whether the sync behind it worked. With Garmin as my only sink, a notification that arrives while the upload is failing looks like success.

I'd like a per-exporter `report_exports: true` (`NTFY_REPORT_EXPORTS` / `TELEGRAM_REPORT_EXPORTS` in env mode, off by default). An exporter with it set runs after the others and appends one line per exporter to the message, like

```
✅ garmin
❌ influxdb: HTTP 401
```

`dispatchExports` already collects this list (`ExportResultDetail[]`) for the ESP32 display, so the notification would just be a second reader of it. With the flag off nothing changes, including timing.

PR follows this issue.

### Use Case

Single user, Garmin is the only sink, phone notification through ntfy. When the Garmin token expires or the upload times out, I want the notification to say so while I'm still in the bathroom. Otherwise I would probably check the app later in the day and see that no weight has been logged in Garmin Connect. It applies to other exporters failing too.

### Alternatives Considered

A separate post-dispatch hook in the orchestrator that posts to a webhook or ntfy. That's a second notification mechanism next to the ntfy and telegram exporters that already exist, with its own config, auth and message format, and the flag on the existing exporters does the same thing with one mechanism. I also didn't want to make every notifier wait unconditionally, because that delays the notification for everyone by the slowest exporter's retries.

alasano:
Closes #343. Branched from #347 because it builds on the shared formatter from that PR, so this one shows both commits until #347 merges and the change here is the second commit.

`report_exports` in YAML, `NTFY_REPORT_EXPORTS` / `TELEGRAM_REPORT_EXPORTS` in env mode, default `false`. The exporter exposes it as `Exporter.reportsExports`. `dispatchExports` splits the eligible exporters on that flag, everything without it runs concurrently as before, then the ones with it run with the collected `ExportResultDetail[]` in `context.exportResults`, and `formatNotification()` appends `✅ name` or `❌ name: error` per entry. `ExportResultDetail` moved from `orchestrator.ts` to `interfaces/exporter.ts` so the context can carry it without importing the orchestrator. The dispatch result is unchanged, success still means at least one exporter succeeded.

With the flag off nothing changes, timing included. With it on, the notification arrives after the slowest exporter and its retries, and the docs say so.

Tests cover the orchestrator ordering and context propagation, all sinks failing still notifies, a notifier-only config, non-reporting exporters not seeing the results, the message formatting in both exporters, env parsing including invalid values, and the registry mapping of the YAML key. Docs and `.env.example` updated. Full suite passing except the two pre-existing macOS-only failures, tsc/eslint/prettier clean.

### DIFF
--- a/.env.example
+++ b/.env.example
@@ -91,6 +91,7 @@ HEIGHT_UNIT=cm
 # NTFY_TOKEN=
 # NTFY_USERNAME=
 # NTFY_PASSWORD=
+# NTFY_REPORT_EXPORTS=false
 
 # ==========================================
 # TELEGRAM CONFIGURATION (only when telegram exporter is enabled)
@@ -99,6 +100,7 @@ HEIGHT_UNIT=cm
 # TELEGRAM_CHAT_ID=987654321
 # TELEGRAM_TITLE=Scale Measurement
 # TELEGRAM_SILENT=false
+# TELEGRAM_REPORT_EXPORTS=false
 
 # ==========================================
 # INTERVALS.ICU CONFIGURATION (only when intervals exporter is enabled)
--- a/docs/exporters.md
+++ b/docs/exporters.md
@@ -187,7 +187,7 @@ global_exporters:
 
 Weight, muscle and bone follow `scale.weight_unit`.
 
-With `report_exports: true` the notification is sent after the other exporters finish and ends with one line per exporter, `✅ garmin` or `❌ garmin: <error>`, so a failed sync is visible on the phone. The notification then arrives once the slowest exporter (and its retries) is done.
+With `report_exports: true` the notification is sent after the other exporters finish and ends with one line per non-reporting exporter, `✅ garmin` or `❌ garmin: <error>`, so a failed sync is visible on the phone. Two notifiers with the flag set do not report on each other. The notification arrives once the slowest exporter (and its retries) is done; with Garmin that can be up to three minutes when its uploader times out and retries. Error text is forwarded as-is (truncated to 120 characters), so a public ntfy topic will carry it.
 
 ## Telegram {#telegram}
 
--- a/docs/exporters.md
+++ b/docs/exporters.md
@@ -167,15 +167,16 @@ global_exporters:
 
 Push notifications to phone/desktop via [ntfy](https://ntfy.sh). Works with ntfy.sh or self-hosted instances.
 
-| Field      | Required | Default             | Description         |
-| ---------- | -------- | ------------------- | ------------------- |
-| `url`      | No       | `https://ntfy.sh`   | Ntfy server URL     |
-| `topic`    | Yes      | (none)              | Topic name          |
-| `title`    | No       | `Scale Measurement` | Notification title  |
-| `priority` | No       | `3`                 | Priority (1 to 5)   |
-| `token`    | No       | (none)              | Bearer token auth   |
-| `username` | No       | (none)              | Basic auth username |
-| `password` | No       | (none)              | Basic auth password |
+| Field            | Required | Default             | Description                         |
+| ---------------- | -------- | ------------------- | ----------------------------------- |
+| `url`            | No       | `https://ntfy.sh`   | Ntfy server URL                     |
+| `topic`          | Yes      | (none)              | Topic name                          |
+| `title`          | No       | `Scale Measurement` | Notification title                  |
+| `priority`       | No       | `3`                 | Priority (1 to 5)                   |
+| `token`          | No       | (none)              | Bearer token auth                   |
+| `username`       | No       | (none)              | Basic auth username                 |
+| `password`       | No       | (none)              | Basic auth password                 |
+| `report_exports` | No       | `false`             | Append the other exporters' results |
 
 ```yaml
 global_exporters:
@@ -186,16 +187,19 @@ global_exporters:
 
 Weight, muscle and bone follow `scale.weight_unit`.
 
+With `report_exports: true` the notification is sent after the other exporters finish and ends with one line per exporter, `✅ garmin` or `❌ garmin: <error>`, so a failed sync is visible on the phone. The notification then arrives once the slowest exporter (and its retries) is done.
+
 ## Telegram {#telegram}
 
 Send a measurement notification to a Telegram chat via a bot. Create a bot with [@BotFather](https://t.me/BotFather) to get a bot token, then start a chat with your bot (or add it to a group/channel) so it can message you.
 
-| Field       | Required | Default             | Description                                    |
-| ----------- | -------- | ------------------- | ---------------------------------------------- |
-| `bot_token` | Yes      | (none)              | Bot token from @BotFather                      |
-| `chat_id`   | Yes      | (none)              | Target chat ID (numeric) or `@channelusername` |
-| `title`     | No       | `Scale Measurement` | First line of the message                      |
-| `silent`    | No       | `false`             | Deliver without a notification sound           |
+| Field            | Required | Default             | Description                                    |
+| ---------------- | -------- | ------------------- | ---------------------------------------------- |
+| `bot_token`      | Yes      | (none)              | Bot token from @BotFather                      |
+| `chat_id`        | Yes      | (none)              | Target chat ID (numeric) or `@channelusername` |
+| `title`          | No       | `Scale Measurement` | First line of the message                      |
+| `silent`         | No       | `false`             | Deliver without a notification sound           |
+| `report_exports` | No       | `false`             | Append the other exporters' results            |
 
 ```yaml
 global_exporters:
@@ -206,7 +210,7 @@ global_exporters:
     silent: false
 ```
 
-The message is sent as plain text. Weight, muscle and bone follow `scale.weight_unit`. In multi-user setups the user's name is prepended as `[Name]`. Historical readings replayed from a scale's offline cache are skipped (a notification for an old measurement is not meaningful).
+The message is sent as plain text. Weight, muscle and bone follow `scale.weight_unit`. `report_exports` works as for [Ntfy](#ntfy). In multi-user setups the user's name is prepended as `[Name]`. Historical readings replayed from a scale's offline cache are skipped (a notification for an old measurement is not meaningful).
 
 ::: tip Finding your chat ID
 Message your bot once, then open `https://api.telegram.org/bot<token>/getUpdates` in a browser — the `chat.id` field holds your chat ID. For groups, add the bot to the group first.
--- a/src/config/env-load.ts
+++ b/src/config/env-load.ts
@@ -62,6 +62,7 @@ export function loadEnvConfig(): AppConfig {
         token: n.token,
         username: n.username,
         password: n.password,
+        report_exports: n.reportExports,
       });
     }
     if (name === 'telegram' && exporterConfig.telegram) {
@@ -71,6 +72,7 @@ export function loadEnvConfig(): AppConfig {
         chat_id: t.chatId,
         title: t.title,
         silent: t.silent,
+        report_exports: t.reportExports,
       });
     }
     if (name === 'intervals' && exporterConfig.intervals) {
--- a/src/exporters/config.ts
+++ b/src/exporters/config.ts
@@ -64,6 +64,7 @@ export interface NtfyConfig {
   token?: string;
   username?: string;
   password?: string;
+  reportExports: boolean;
 }
 
 export interface FileConfig {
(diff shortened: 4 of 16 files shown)


KristianP26 (changes requested):
Adds an opt-in `report_exports` flag to the ntfy and telegram exporters: `dispatchExports` now splits the eligible exporters on a new `Exporter.reportsExports` flag, runs the non-reporting ones concurrently exactly as before, then runs the reporters with the collected `ExportResultDetail[]` in `context.exportResults`, and `formatNotification()` appends one line per outcome. The ordering guarantee is sound: the split is on the flag, not on array position, so a notifier listed first in `global_exporters` still runs in the second wave, there is no cycle and therefore no deadlock, and `[...details]` is copied before the reporter results are pushed back. Exit-code semantics are preserved exactly: the old `allFailed` accumulator over one wave is replaced by `details.every((d) => !d.ok)` over the concatenation of both waves, which is the same predicate, and `src/index.ts:329` still exits 1 only when nothing succeeded. The secrets concern does not materialise on any current exporter (verified empirically, see below), but one real blocker remains: env mode advertises `NTFY_REPORT_EXPORTS` / `TELEGRAM_REPORT_EXPORTS` and validates them, yet never wires them through to the exporter, so they are silently inert.

## Findings

### 1. Blocker: NTFY_REPORT_EXPORTS / TELEGRAM_REPORT_EXPORTS are parsed and validated but never reach the exporter
`src/config/env-load.ts:55`

The PR adds `reportExports` to `NtfyConfig`/`TelegramConfig` and parses `NTFY_REPORT_EXPORTS` / `TELEGRAM_REPORT_EXPORTS` in `src/exporters/config.ts:252` and `:312`, documents both in `.env.example`, and adds tests asserting they parse (including throwing on `maybe`). But the env-mode runtime path is `loadAppConfig` -> `loadEnvConfig()` (`src/config/load.ts:50`), which rebuilds each exporter as a YAML-shaped `ExporterEntry` in `src/config/env-load.ts`. The ntfy block (lines 55 to 66) maps url/topic/title/priority/token/username/password and the telegram block (lines 67 to 74) maps bot_token/chat_id/title/silent; neither maps `report_exports`. `createExporterFromEntry` then reads `config.report_exports` (`src/exporters/registry.ts:115`, `:149`), finds nothing, and defaults to false. `createExporters(config: ExporterConfig)` in `src/exporters/index.ts:18`, which does consume the parsed `NtfyConfig` directly, is dead code, referenced only by `tests/exporters/index.test.ts`. So the whole env-mode half of the feature is a no-op, and the test suite cannot catch it because `tests/exporters/config.test.ts` stops at `loadExporterConfig()` and never crosses `loadEnvConfig()`.

**What goes wrong:** A Docker/.env user (no config.yaml) sets `EXPORTERS=ntfy`, `NTFY_TOPIC=my-scale`, `NTFY_REPORT_EXPORTS=true`. I ran this through the real code path on the PR branch: `loadEnvConfig()` produces `{"type":"ntfy","url":"https://ntfy.sh","topic":"my-scale","title":"Scale Measurement","priority":3}` with no `report_exports` key, and `createExporterFromEntry(entry).reportsExports === false`. The notification arrives with no exporter result lines, the user has no error and no warning, and `NTFY_REPORT_EXPORTS=maybe` still throws, which makes the variable look wired up.

**Suggested fix:** Add `report_exports: n.reportExports` to the ntfy `Object.assign` in `src/config/env-load.ts` and `report_exports: t.reportExports` to the telegram one. Cover it with a test that goes through `loadEnvConfig()` (or through `createExporterFromEntry(loadEnvConfig().global_exporters[0])`) rather than stopping at `loadExporterConfig()`, since that is the seam the current tests miss.

### 2. Minor: "one line per exporter" is wrong when both notification exporters set report_exports
`docs/exporters.md:190`

Both reporters run in the same second wave and are handed `[...details]` captured from the first wave only (`src/orchestrator.ts`), so with `report_exports: true` on ntfy and telegram at once, neither message lists the other. The docs promise "one line per exporter", which reads as all of them. The same paragraph says the notification "arrives once the slowest exporter (and its retries) is done" without a magnitude; with Garmin that is `SUBPROCESS_TIMEOUT_MS = 60_000` (`src/exporters/garmin.ts:16`) times the default 3 attempts from `withRetry` (`src/utils/retry.ts:50`), so up to about 3 minutes of phone silence after stepping off the scale.

**What goes wrong:** User enables `report_exports` on both ntfy and telegram plus garmin and influxdb. Expected from the docs: four lines in each message. Actual: two lines in each (garmin, influxdb), with telegram absent from the ntfy message and vice versa. Separately, Garmin's Python uploader hangs; the user reads "after the other exporters finish" and assumes seconds, then sees no notification for three minutes and assumes the sync broke.

**Suggested fix:** Reword to "one line per non-reporting exporter" (and note that two notifiers with the flag do not report on each other), and add the worst case, e.g. "with Garmin this can be up to three minutes when its uploader times out and retries".

### 3. Minor: Exporter error text is forwarded verbatim to a channel that is public by default
`src/exporters/notification-message.ts:20`

`lines.push(... : `❌ ${r.name}: ${r.error}`)` copies each failing exporter's raw message into the notification body. I checked every current error source and found no credential leak: `httpError` (`src/utils/retry.ts:35`) emits only `HTTP <status>`; undici gives `fetch failed` with the URL only on the unread `cause` (verified by running fetch against a host with a token in the path); and mqtt.js on a `mqtt://user:SUPERSECRET@host` URL gives `connect ECONNREFUSED 127.0.0.1:1` / `getaddrinfo ENOTFOUND ...` / `connack timeout`, never the credentials (also verified). What does travel is local filesystem detail and third-party exception text: `FileExporter` returns `errMsg(err)` for fs failures (`src/exporters/file.ts:96`), which is Node's `ENOENT: no such file or directory, open '<absolute path>'`; Strava throws `Strava token file not found at <path>` (`src/exporters/strava.ts:87`); and Garmin surfaces the Python `str(e)` verbatim (`garmin-scripts/garmin_upload.py:129`). ntfy.sh topics are unauthenticated by default, so anyone who knows the topic name reads all of it. There is also no length bound on the appended block.

**What goes wrong:** User exports to a public ntfy topic plus a `file` exporter at `/mnt/nas/health/readings.csv`. The NAS unmounts, the write fails, and the notification posted to the public topic reads `❌ file: ENOENT: no such file or directory, open '/mnt/nas/health/readings.csv'`, disclosing the host's directory layout to anyone subscribed to the topic.

**Suggested fix:** Truncate each forwarded error to a sane length (say 120 chars) in `formatNotification`, and add a one-line caution to the `report_exports` docs that exporter error text is forwarded as-is, so a public ntfy topic will carry it. A future exporter that does embed a URL or token in its message would otherwise inherit the leak silently.

### 4. Minor: The new field's description is never shown by the wizard, so the delay trade-off is invisible during setup
`src/exporters/ntfy.ts:48`

The schema entry sets `description: 'Wait for the other exporters and append their results to the message'`, but `promptField` in `src/wizard/steps/exporters.ts` only reads `field.description` in the `password` branch (line 30, for the `${ENV_VAR}` hint). The `boolean` branch at line 59 renders `prompts.confirm(`${field.label}?`)` and nothing else, so `npm run setup` asks a bare "Report export results?" This is a pre-existing wizard limitation, but it bites harder here than for `silent` or `retain` because this flag serialises the notification behind every other exporter.

**What goes wrong:** User runs `npm run setup`, is asked "Report export results?", answers yes assuming it is a cosmetic addition, and later cannot explain why the phone notification now lags the weigh-in by up to three minutes whenever Garmin is slow.

**Suggested fix:** Fold the trade-off into the `label` for this field, e.g. `label: 'Report export results (waits for the other exporters)'`, in both `ntfySchema` and `telegramSchema`.

### 5. Minor: Merge-order hazard: this PR carries #347's commit and the repo is squash-only
`src/exporters/notification-message.ts:1`

`git log origin/dev..` on the head shows two commits: 7693d65 (the #347 weight-unit formatter, which creates `src/exporters/notification-message.ts` and rewrites the ntfy/telegram `formatMessage`) and 3763a8a (this feature). `gh api repos/:owner/:repo` reports `allow_squash_merge: true`, `allow_merge_commit: false`, `allow_rebase_merge: false`, so #347 can only land as a squash with a new SHA. Once that happens, 7693d65 is no longer an ancestor of dev and this branch conflicts on the overlap: an add/add conflict on `src/exporters/notification-message.ts` plus content conflicts in `src/exporters/ntfy.ts`, `src/exporters/telegram.ts`, `src/interfaces/exporter.ts`, `src/runtime/processor.ts`, `tests/exporters/ntfy.test.ts`, `tests/exporters/telegram.test.ts` and `tests/runtime/processor.test.ts`. GitHub currently reports MERGEABLE/CLEAN only because #347 has not landed yet.

**What goes wrong:** Maintainer squash-merges #347 first. This PR immediately flips to CONFLICTING across seven files, and the contributor has to rebase and drop the duplicated commit before it can land.

**Suggested fix:** Merge #350 on its own and close #347 as contained in it (the tree is identical either way), or merge #347 first and ask the contributor to rebase. Do not squash #347 and then expect this branch to still apply.

## Worth calling out

- **The ordering test is genuinely order-sensitive, and the allFailed refactor is provably behaviour-preserving** (tests/orchestrator.test.ts:222) The "runs reporting exporters after the others and hands them the outcomes" test uses a garmin stub that sleeps 50ms before pushing to `order`, then asserts `order` equals `['garmin', 'ntfy']`. Under the old single-wave `Promise.allSettled` the synchronous ntfy stub would resolve first and the array would read `['ntfy', 'garmin']`, so the test really does fail without the change rather than passing vacuously. The refactor of `allFailed` from a loop-carried flag to `details.every((d) => !d.ok)` over the concatenated waves is the same predicate over the same set, which is what keeps `src/index.ts:329` (`if (!success) process.exit(1)`) and the continuous-mode path unchanged, and "still notifies when every other exporter failed" plus "returns success false when the others and the reporter all fail" pin both sides of that boundary.

## Questions for you

- Please wire `report_exports` through `src/config/env-load.ts` for both ntfy and telegram, and add a test that crosses `loadEnvConfig()` rather than stopping at `loadExporterConfig()`. Right now the two env vars in `.env.example` are inert.

- Confirm the intended behaviour when both ntfy and telegram set `report_exports: true`. Neither currently reports on the other, which may be fine, but the docs should say so.

- Which way do you want the #347 overlap resolved? Merging #350 alone lands both commits; squash-merging #347 first will force a rebase here.

alasano:
The env path did nothing, you were right. `report_exports` now gets copied in `loadEnvConfig()` for both ntfy and telegram, and the new test goes through `loadAppConfig()` down to the built exporter's `reportsExports` instead of stopping at `loadExporterConfig()`. It fails if either mapping line is removed.

The rest is in too. Forwarded error text is truncated to 120 characters, sliced by code point so a multibyte character can't get cut into a replacement char. The wizard label now reads "Report export results (waits for the other exporters)" so the delay is visible during setup. The docs say one line per non-reporting exporter, that two notifiers with the flag don't report on each other, and that Garmin can push the notification out by up to three minutes.

Two notifiers not reporting on each other is intended, and the docs now say so.

For the overlap with the weight-unit PR, merge that one first since it's already approved, and I'll rebase this onto dev right after, which drops it to a single commit.

KristianP26:
`#347` is merged, so this is clear to rebase. Its commit is on `dev` now as a squash, so the copy on this branch will conflict until you drop it; after the rebase this should come down to the single `feat(notifications)` commit plus the env wiring.

The env-path fix and the truncation are what I asked for and I will re-verify both on the rebased head.

KristianP26 (approved):
Approved and merging. I rebased this onto `dev` myself rather than leaving it parked, since `#347` had landed and the copy of its commit on this branch was the only thing conflicting. Nothing of yours changed: the rebase dropped that one commit and replayed your two.

The env-path fix is verified rather than read. Deleting either mapping line in `loadEnvConfig()` fails the new test (`expected false to be true`), so the whole `loadAppConfig` chain down to the built exporter's `reportsExports` is genuinely pinned now, which is what the previous version could not catch.

Full suite on the rebased head: 133 files, 2278 tests, tsc, eslint and prettier clean.

Truncating forwarded error text by code point rather than by byte is the right call, and the wizard label naming the delay is what stops this reading as a bug the first time Garmin holds a notification back.

Issue #346, opened 2026-08-23
mqtt-proxy: 60 s reading timeout counts from connect, so an autonomous connect that precedes the weigh-in kills the session mid-reading


alasano:
### Description

With `auto_connect` the ESP32 connects the instant the scale advertises, and a QN scale keeps advertising for a while after any wake-up, so the connect often happens well before anyone stands on it. The autonomous path in `src/ble/handler-mqtt-proxy/watcher.ts` wraps the whole session in `withTimeout(pending, 60_000)` counted from that connect. Today I stepped on at T+54 s, the scale streamed its unstable 0x10 frames for 2.5 s and went quiet for its impedance phase, and at T+60 s the host tore the session down with `GATT reading timeout (autonomous)`, a few seconds before the stable frame would have come. The scale finished the measurement on its own and showed the weight, the host never got it. The host-initiated mqtt-proxy path has the same fixed 60 s, and the native paths use `RAW_READING_TIMEOUT_MS` / `session_timeout_sec` the same way, from connect.

After the teardown the scale re-advertises, the ESP32 reconnects within about 10 s, the scale never answers 0x13 in that state, and the session times out again at 60 s. That loop ran for five minutes after I had left the room with the scale's display off.

The fix I went with is to measure the deadline as silence, restarting it on every notification from the scale and not on outgoing writes, so a silent scale still times out. It has to restart before the adapter gate, because the unstable 0x10 frames are rejected by the adapter and they are the proof that a weigh-in is in progress. The values stay what they are, 60 s on mqtt-proxy and `session_timeout_sec` on native, only what they count changes. PR follows this issue.

### Steps to Reproduce

1. ES-CS20M on `handler: mqtt-proxy` with `auto_connect`, debug logs on
2. Wake the scale (tap it) without standing on it, the ESP32 connects
3. Step on 50+ s later and stay on until the scale shows the final weight
4. Host logs `GATT reading timeout (autonomous)` 60 s after the connect, during the scale's silent impedance phase

### Scale Brand / Model

Renpho ES-CS20M (QN classic dialect, proto 0x15)

### Platform

Docker (Linux, mqtt-proxy via ESP32-S3, autonomous connect)

### Node.js Version

image default (v1.24.0 container)

### App Version

1.24.0 with #337 and #339 applied

### Debug Logs

Connect at 15:14:56, weigh-in starts at 15:15:50, session killed at 15:15:56 (annotations mine):

```
[...]

Issue #296, opened 2026-07-26
"mqtt-proxy: autonomous connected event published but not consumed by ble-scale-sync; ESP32 scan doesn't resume after connect"


vanboxel:
### Description

Using v1.21.1 (HA add-on) with the official, freshly-flashed ESP32-S3 mqtt-proxy firmware (no local modifications) against a Robi S9 (MAC 50:FB:19:DF:6F:6D, GATT-only, addr_type=0).

Symptom: after a fresh ESP32 reset, stepping on the scale triggers a successful autonomous GATT connect on the firmware side within 750–1500ms. The `connected` payload is published to `ble-proxy/esp32-scale-proxy/connected` and is confirmed received by an independent `mosquitto_sub` listener (full 17-char payload, `"autonomous": true`). However, ble-scale-sync itself does not appear to consume this message in most attempts — no `Matched`/`connected`/measurement lines appear in its log, and no export happens. The add-on continues its normal scan_cooldown restart cycle as if nothing happened.

Additional observations:
- The ESP32 must be freshly reset immediately before each measurement attempt. Without a reset, no autonomous connect is ever triggered, even when the scale is woken.
- After a successful autonomous connect, the ESP32's own streaming scan loop does not appear to resume (no further "Streaming scan: N devices" lines in serial output afterward), so only one connect attempt is possible per ESP32 boot.
- Confirmed the `config` topic correctly reaches the ESP32 (`Config: 1 scale MAC(s), autoConnect=True, lazyNotify=True` appears at boot), so the scale MAC is known.

Setup:
```yaml
ble:
  handler: mqtt-proxy
  mqtt_proxy:
    broker_url: 'mqtt://192.168.0.143:1883'
    device_id: esp32-scale-proxy
    topic_prefix: ble-proxy
    username: blescale
    password: 'BleScale123!'
```

Repro rate: roughly 1-in-several attempts results in ble-scale-sync actually processing the measurement; most resets+steps produce no reaction at all in the add-on log despite the confirmed MQTT delivery.

Question: is there a known timing window where ble-scale-sync's own scan/restart cycle needs to be actively subscribed for the autonomous `connected` event to be picked up? And separately — is it expected that the ESP32 streaming scan does not resume after one autonomous connect, requiring a manual reset before the next attempt?

### Steps to Reproduce

[...]

Issue #291, opened 2026-07-21
Hutbit (#254/#268/#278): matching resolves to Hutbit and the scale streams FFB2 notifications over the ESPHome proxy, but the add-on decodes zero frames


bondesen:
Follow-up to #278 (I'm the Hutbit reporter/contributor). With the SWAN manufacturer-signature match from #279 in place, matching now works correctly — the unit resolves to **Hutbit**, not MGB/Robi, and the link no longer drops. But the scale still never produces a reading, and live testing tonight isolates the failure one layer deeper: **the proxy receives a continuous FFB2 notification stream from the scale, but the add-on never decodes a single frame.**

## Setup

- Unit: model 218008 / WL292, advertises `Complete Local Name: SWAN`, identity MAC `03:B3:EC:93:B8:7E` (the #254/#278 device).
- Transport: `handler: esphome-proxy`, pointed at a **dedicated, non-adopted** ESPHome bluetooth-proxy (ESP32, ESPHome 2026.7.1, `bluetooth_proxy: active: true`, active BLE scan). HA never adopts it, so it owns the advertisement subscription.
- BLE Scale Sync: HA add-on, current build (matching resolves to Hutbit, so the #279 signature match is present), `custom_config: true`.

## Symptom — add-on side (BLE Scale Sync)

```
[BLE] ESPHome proxy connected at bt-proxy-vaegt.local:6053
[BLE] ESPHome proxy transport ready (broadcast + GATT). ... Hutbit ...
[BLE] ESPHome ReadingWatcher started, listening for advertisements
[BLE] Matched: Hutbit (03:B3:EC:93:B8:7E); opening GATT via ESPHome proxy
[BLE:debug] Multi-char mode: 2 bindings
[BLE:debug] Subscribing to 0000ffb200001000800000805f9b34fb (notify)...
[BLE] Subscribed to 1 notification(s). Step on the scale.
[BLE:debug] Calling adapter.onConnected()
[BLE:debug] Hutbit: handshake sent
[BLE:debug] adapter.onConnected() completed
   <-- and then nothing. No "Hutbit frame:", no "Reading complete", ever. -->
```

A full weigh-in — standing until the display auto-powered off (~1 minute), BT-connected symbol lit on the scale the whole time — produced **zero** `Hutbit frame:` debug lines and no reading. Repeated across several restarts.

## The new evidence — proxy side (ESP32, VERBOSE)

[...]

Issue #281, opened 2026-07-16
Proxy transports have no liveness watchdog: a wedged mqtt-proxy/esphome-proxy link is indistinguishable from idle


KristianP26:
## Problem

`buildReadingSource` (`src/runtime/sources.ts:53-65`) returns the proxy `ReadingWatcher` with **neither** a `ConsecutiveFailureWatchdog` **nor** an `onFailure` hook. The native poll path gets both (`sources.ts:74-86`). So on `mqtt-proxy` and `esphome-proxy` there is no in-app liveness recovery at all: if the proxy link wedges, `nextReading()` simply never resolves and the app waits forever.

## Why it matters now

Until #277 this was accidentally masked on HAOS. The stale heartbeat flipped the container to `unhealthy` roughly every 7 minutes and the Supervisor watchdog restarted it, which incidentally recovered a wedged proxy. That was never a liveness mechanism (it fired on healthy idle too, which is the bug #277 fixed), but it did paper over this gap. With the heartbeat now on a timer, an idle-looking-but-wedged proxy stays wedged.

Partial mitigation already exists: `handler-esphome-proxy/client.ts:172` sets `reconnect: true` and subscriptions restore after reconnect, so the common dead-link case self-heals. This issue is about the cases that do not.

## The hard part

A wedged proxy and a normal idle scale are **indistinguishable** from the loop's point of view: in both, `nextReading()` never returns. So a plain timeout is wrong, it would restart-loop anyone who does not weigh in.

A workable signal: advertisements flow constantly while the transport is alive, independent of weigh-ins. The native path already has this concept (`LIVENESS_PROBE_WINDOW_MS`, `src/ble/types.ts:75`) which distinguishes "radio hears nothing at all" from "the scale is just not being stood on". The proxy watcher could apply the same idea: if the transport has delivered zero advertisements from any device for N minutes, the link is wedged rather than idle.

## Acceptance

- Wedged proxy (no adverts at all) trips the watchdog and exits so the supervisor restarts.
- A genuinely idle scale (adverts flowing from other devices, nobody weighing in) never trips it.

Found while fixing #277.

KristianP26:
Status update: half of this is now handled, and it is worth recording which half so the remaining work is not re-scoped from scratch.

[...]

PR #351, opened 2026-08-23
fix(ble): measure the reading deadline as scale silence, not time since connect

alasano:
Full log in #346. Branched from #339 because it rewrites the same two watcher call sites that PR touches, so this one shows both commits until #339 merges and the change here is the second commit.

`withIdleTimeout()` sits next to `withTimeout()` in `src/ble/types.ts` and restarts its timer whenever the callback it hands the caller is invoked. `waitForRawReading()` takes that callback as `onActivity` and calls it first thing in `handleNotification`, before the ACK and the adapter parse, so a frame the adapter rejects (an unstable QN 0x10) still counts. Outgoing writes don't, so a scale that says nothing still times out on schedule. All four reading call sites switched, both mqtt-proxy paths keep 60 s (now a named `GATT_READING_IDLE_MS`) and noble and node-ble keep `readingTimeoutMs ?? RAW_READING_TIMEOUT_MS`. The `pending.catch(() => {})` in the watcher went away because the helper races the promise itself, and the `fireDisconnect()` then `cleanup()` teardown from #339 is untouched.

The `session_timeout_sec` row and tip in the docs now describe silence, and the comments on the constants match.

Two unit tests on the helper (silence rejects, activity restarts) and one autonomous regression in `handler-mqtt-proxy.test.ts`, idle for 55 s, a frame the adapter rejects, past 60 s no disconnect is published, a complete frame at 61 s resolves. That test fails on the previous code, the disconnect is published at 60 s, and passes with the change. The leak test from #339 still passes, pure silence tears the session down at 60 s. Full suite passing except the two pre-existing macOS-only failures, tsc/eslint/prettier clean.

### DIFF
--- a/docs/guide/configuration.md
+++ b/docs/guide/configuration.md
@@ -71,7 +71,7 @@ ble:
 | `noble_driver`        | No                          | OS default     | `abandonware` or `stoprocent`. Overrides the default BLE driver. Only applies when `handler: auto`.                                                                                                                                                |
 | `adapter`             | No                          | System default | Linux only. Select a specific Bluetooth adapter (e.g., `hci0`, `hci1`). See below.                                                                                                                                                                 |
 | `force_scale_adapter` | No                          | Auto-detect    | Name of the scale protocol adapter to use, bypassing auto-detection. Requires `scale_mac`. See below.                                                                                                                                              |
-| `session_timeout_sec` | No                          | `120`          | Seconds one GATT session may wait for a complete reading (5 to 600). Native BLE handlers only; ignored on `mqtt-proxy` and `esphome-proxy`. See below.                                                                                             |
+| `session_timeout_sec` | No                          | `120`          | Seconds of scale silence that end a GATT session (5 to 600); an inbound frame restarts the clock. Native BLE handlers only; ignored on `mqtt-proxy` and `esphome-proxy`. See below.                                                                |
 | `qn_protocol_byte`    | No                          | Auto           | QN-family scales only. Protocol byte the handshake echoes back to the scale (0 to 255). Set it when a QN scale runs the whole handshake and then reports nothing, or when its scale-info frame is lost in transit on a proxy transport. See below. |
 | `qn_report_byte`      | No                          | `254` (0xFE)   | QN-family scales only. Payload byte of the history-response frame (0 to 255). Try `252` (0xFC) when a QN scale completes the handshake and then reports nothing. See below.                                                                        |
 | `mqtt_proxy`          | If `handler: mqtt-proxy`    | (none)         | MQTT proxy connection (`broker_url`, `device_id`, `topic_prefix`, `username`, `password`, `auto_connect`, `embedded_broker_*`). See [ESP32 BLE Proxy](./esp32-proxy).                                                                              |
@@ -153,7 +153,7 @@ If `252` makes your scale produce a weight, please say so in an issue with the m
 ::: tip Shortening the session (`session_timeout_sec`)
 Some scales will not run a standalone weigh-in while a host holds the GATT session open. The Beurer BF500 is the clearest example: it displays `APP` and waits, so only a measurement taken **between** sessions is picked up.
 
-By default a session waits 120 seconds for a reading. On a scale like this, that is 120 seconds out of every cycle in which stepping on it achieves nothing. Shortening the session, and lengthening the gap after it, frees the scale for most of the cycle:
+By default a session ends after 120 seconds without a notification from the scale. On a scale like this, that is 120 seconds out of every cycle in which stepping on it achieves nothing. Shortening the session, and lengthening the gap after it, frees the scale for most of the cycle:
 
 ```yaml
 ble:
--- a/src/ble/handler-mqtt-proxy/watcher.ts
+++ b/src/ble/handler-mqtt-proxy/watcher.ts
@@ -5,7 +5,7 @@ import { waitForRawReading } from '../shared.js';
 import { resolveAdapter } from '../../scales/resolve.js';
 import { evaluateAdvertisement, GraceTimers, DedupWindow, logAdvert } from '../advertisement.js';
 import type { Watcher, WatcherConfig } from '../reading-source.js';
-import { bleLog, withIdleTimeout, errMsg, IMPEDANCE_GRACE_MS } from '../types.js';
+import { bleLog, withIdleTimeout, withTimeout, errMsg, IMPEDANCE_GRACE_MS } from '../types.js';
 import { AsyncQueue } from '../async-queue.js';
 import { topics } from './topics.js';
 import {
@@ -27,6 +27,14 @@ import { type ScanResultEntry, toBleDeviceInfo } from './scan.js';
 const DEDUP_WINDOW_MS = 30_000;
 /** Milliseconds of scale silence that end a GATT reading session. */
 const GATT_READING_IDLE_MS = 60_000;
+/**
+ * Absolute cap on one GATT session regardless of activity. A scale that keeps
+ * streaming adapter-rejected frames restarts the idle timer forever, and while
+ * the session is up the ESP32 holds its single connection and pauses scanning.
+ * Kept at ReadingWatcher.GATT_STALE_MS so the stale reset's assumption that no
+ * session outlives it stays true.
+ */
+const GATT_SESSION_ABSOLUTE_MS = 90_000;
 
 /** Bluetooth Base UUID for expanding 16-bit UUIDs to 128-bit form. */
 const BT_BASE_UUID = '00000000-0000-1000-8000-00805f9b34fb';
@@ -471,21 +479,25 @@ export class ReadingWatcher implements Watcher {
         entry.address,
       );
       bleLog.debug(`GATT read driven by adapter: ${gattAdapter.name} (${entry.address})`);
-      const raw = await withIdleTimeout(
-        (onActivity) =>
-          waitForRawReading(
-            connected.charMap,
-            connected.device,
-            gattAdapter,
-            profile,
-            entry.address.replace(/[:-]/g, '').toUpperCase(),
-            undefined,
-            undefined,
-            undefined,
-            onActivity,
-          ),
-        GATT_READING_IDLE_MS,
-        `GATT reading timeout for ${entry.address}`,
+      const raw = await withTimeout(
+        withIdleTimeout(
+          (onActivity) =>
+            waitForRawReading(
+              connected.charMap,
+              connected.device,
+              gattAdapter,
+              profile,
+              entry.address.replace(/[:-]/g, '').toUpperCase(),
+              undefined,
+              undefined,
+              undefined,
+              onActivity,
+            ),
+          GATT_READING_IDLE_MS,
+          `GATT reading timeout for ${entry.address}`,
+        ),
+        GATT_SESSION_ABSOLUTE_MS,
+        `GATT session cap exceeded for ${entry.address}`,
       );
       registerScaleMac(this.config, entry.address).catch(() => {});
       this.queue.push(raw);
@@ -605,21 +617,25 @@ export class ReadingWatcher implements Watcher {
         `Autonomous connect: charMap built with ${charMap.size} chars, waiting for reading...`,
       );
 
-      const raw = await withIdleTimeout(
-        (onActivity) =>
-          waitForRawReading(
-            charMap,
-            dev,
-            adapter,
-            profile,
-            data.address.replace(/[:-]/g, '').toUpperCase(),
-            undefined,
-            undefined,
-            undefined,
-            onActivity,
-          ),
-        GATT_READING_IDLE_MS,
-        `GATT reading timeout for ${data.address} (autonomous)`,
+      const raw = await withTimeout(
+        withIdleTimeout(
+          (onActivity) =>
+            waitForRawReading(
+              charMap,
+              dev,
+              adapter,
+              profile,
+              data.address.replace(/[:-]/g, '').toUpperCase(),
+              undefined,
+              undefined,
+              undefined,
+              onActivity,
+            ),
+          GATT_READING_IDLE_MS,
+          `GATT reading timeout for ${data.address} (autonomous)`,
+        ),
+        GATT_SESSION_ABSOLUTE_MS,
+        `GATT session cap exceeded for ${data.address} (autonomous)`,
       );
       registerScaleMac(this.config, data.address).catch(() => {});
       bleLog.info(
--- a/src/ble/handler-mqtt-proxy/watcher.ts
+++ b/src/ble/handler-mqtt-proxy/watcher.ts
@@ -25,7 +25,7 @@ import { registerScaleMac, publishConfig } from './display.js';
 import { type ScanResultEntry, toBleDeviceInfo } from './scan.js';
 
 const DEDUP_WINDOW_MS = 30_000;
-/** Seconds of scale silence that end a GATT reading session. */
+/** Milliseconds of scale silence that end a GATT reading session. */
 const GATT_READING_IDLE_MS = 60_000;
 
 /** Bluetooth Base UUID for expanding 16-bit UUIDs to 128-bit form. */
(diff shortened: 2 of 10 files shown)


KristianP26 (changes requested):
The change replaces the fixed reading deadline with `withIdleTimeout()`, a timer that restarts on every inbound notification, and wires it into the four bounded `waitForRawReading()` call sites (noble, node-ble, and both mqtt-proxy watcher paths). The seam is chosen correctly: `onActivity` fires in `shared.ts:472` after the `if (resolved) return` guard but before the ACK and the adapter parse, so a frame the adapter rejects still counts, while outgoing writes do not, and the broadcast-side `GraceTimers` and the GATT-side `HoldTimer` are untouched, so nothing double-counts. The core fix is right and the mqtt-proxy regression test genuinely fails on the old code. What is missing is an absolute ceiling: on the mqtt-proxy path the host used to guarantee the ESP32 was released within 30 s connect + 60 s read, and now guarantees nothing, leaving the only cap in the separately flashed firmware. Three comments (the Zod schema, the `POLL_CYCLE_TIMEOUT_MS` arithmetic, and the `RAW_READING_TIMEOUT_MS` rationale) now contradict the new semantics. Note the branch also carries PR #339's commit 81e418a, which is out of scope here.

## Findings

### 1. Major: No host-side absolute cap on a GATT session; the mqtt-proxy ceiling now lives only in the firmware
`src/ble/handler-mqtt-proxy/watcher.ts:487`

Before this change the host bounded a proxy GATT session at COMMAND_TIMEOUT_MS (30 s, see the `advanceTimersByTimeAsync(30_000)` connect-timeout test) plus the fixed 60 s read, i.e. exactly ReadingWatcher.GATT_STALE_MS (watcher.ts:361). With the idle timer, any stream of frames arriving more often than every 60 s keeps the session alive with no upper bound anywhere in the host process: `src/runtime/loop.ts:68` calls `source.nextReading(signal)` with no timeout wrapper (watcher sources wait indefinitely by design, cf. owner issue #281), and `POLL_CYCLE_TIMEOUT_MS` only guards the native poll path in `src/runtime/poll-source.ts:57`, never the watcher. The only remaining cap is `GATT_SESSION_MAX_MS = 180000` in firmware/main.py:45, which is a separately flashed codebase: the guard only landed with #296 (v1.21.2), it is `getattr(board, ...)`-overridable, and anything older has no cap at all. While the session is up the firmware holds `_scan_paused` (firmware/main.py:366 and 448) and its single BLE connection, so the whole watcher is deaf for the duration. On the native path the effective ceiling silently moves from `readingTimeoutMs` to POLL_CYCLE_TIMEOUT_MS (900 s), 7.5x the documented 120 s.

**What goes wrong:** mqtt-proxy continuous mode, ESP32 running pre-#296 firmware (or a board that raised GATT_SESSION_MAX_MS). A scale that streams frames the adapter rejects at least once a minute (the zeroed body-comp stubs of #229, the all-zero Hutbit frames of #291, or a QN streaming unstable 0x10) keeps calling `onActivity`, so `withIdleTimeout` never rejects. `handleGattReading` never returns, `gattInProgress` stays true, the firmware never resumes scanning, and the watcher produces no reading and no error at all. Previously the host tore the session down at 60 s and the proxy went back to scanning.

**Suggested fix:** Keep the idle timer but nest it in an absolute one, e.g. `await withTimeout(withIdleTimeout((onActivity) => waitForRawReading(...), GATT_READING_IDLE_MS, msg), GATT_SESSION_ABSOLUTE_MS, msg)`. On the proxy paths pick a value at or under `ReadingWatcher.GATT_STALE_MS` so the host keeps the invariant its own stale-reset assumes; on the native paths a documented multiple of `readingTimeoutMs` (or `POLL_CYCLE_TIMEOUT_MS` minus the connect budget) keeps the `session_timeout_sec` contract meaningful.

### 2. Minor: POLL_CYCLE_TIMEOUT_MS rationale still budgets a bounded 120 s reading phase
`src/ble/types.ts:51`

The comment reads "The worst legitimate node-ble cycle is roughly 575 s (discovery 120 + six connect attempts 170 + GATT acquisition 30 + characteristic retries + reading 120), so 900 s never fires on a healthy run." After this PR the reading phase has no upper bound at all, so the "+ reading 120" term and the "never fires on a healthy run" conclusion are both false. This is the one comment in the file that reasons about the reading phase as a fixed budget, and the PR updated the other two constants' comments but not this one.

**What goes wrong:** Native handler, continuous mode, a scale that keeps emitting adapter-rejected frames for 15 minutes. `withIdleTimeout` never fires, `POLL_CYCLE_TIMEOUT_MS` does, and the operator sees "Scan cycle exceeded 900s and was abandoned (transport wedge?)" instead of "Timed out waiting for a complete scale reading". They chase a nonexistent D-Bus wedge, and the cycle counts toward `watchdog_max_consecutive_failures`.

**Suggested fix:** Rewrite the arithmetic to name the reading phase as unbounded-while-frames-arrive, or add the absolute cap from the MAJOR above and keep the budget honest by citing that constant instead of 120.

### 3. Minor: Zod schema still documents session_timeout_sec with the old deadline semantics
`src/config/schema.ts:124`

`docs/guide/configuration.md:73` and `:118` and `src/ble/types.ts:123` were all rewritten to say silence, but the doc comment on the option itself still reads "How long one GATT session may wait for a complete reading, in seconds (default 120)." `src/config/schema.ts` is the canonical definition of the user-facing key, so the PR leaves the two authoritative descriptions of the same option contradicting each other.

**What goes wrong:** A contributor (or the reporter of #83) reads the schema to work out why a `session_timeout_sec: 20` scale is held for longer than 20 s, concludes from this comment that the option is broken, and files a bug against behaviour that is now intentional.

**Suggested fix:** Update the comment to match the new wording, e.g. "Seconds of scale silence that end one GATT session (default 120); every notification restarts the clock." Keep the two cost sentences about adapter resets and the failure watchdog, they are still accurate.

### 4. Minor: RAW_READING_TIMEOUT_MS justified by a transport that never uses it
`src/ble/types.ts:25`

The new comment says "The ESP32 proxy connects as soon as the scale advertises, often well before anyone steps on, and a window counted from connect ran out during a weigh-in." `RAW_READING_TIMEOUT_MS` is read only at `src/ble/handler-noble-shared.ts:615` and `src/ble/handler-node-ble/scan.ts:443`, both native handlers. The mqtt-proxy uses its own `GATT_READING_IDLE_MS = 60_000` (handler-mqtt-proxy/watcher.ts:29) and the esphome-proxy uses neither. So the motivating evidence is attached to the one constant the reported case cannot reach.

**What goes wrong:** Someone tuning the native reading window reads this and assumes proxy sessions honour the same 120 s constant, then cannot explain why an mqtt-proxy session ends after 60 s of silence.

**Suggested fix:** State the mechanism generically on `RAW_READING_TIMEOUT_MS` (a window counted from connect can expire before the user steps on, since connect is driven by the advertisement, not the weigh-in) and move the ESP32-specific narrative to `GATT_READING_IDLE_MS`, which is the constant that case actually exercised.

### 5. Minor: GATT_READING_IDLE_MS doc says "Seconds" for a millisecond constant
`src/ble/handler-mqtt-proxy/watcher.ts:28`

`/** Seconds of scale silence that end a GATT reading session. */ const GATT_READING_IDLE_MS = 60_000;` The name carries `_MS` and the value is milliseconds, so the unit word in the comment is wrong. The neighbouring `DEDUP_WINDOW_MS = 30_000` has no unit word at all, so there is no precedent for reading the comment as seconds-of-wall-clock prose.

**What goes wrong:** A future edit trusts the comment, reads 60_000 as seconds, and "corrects" the constant to 60 to match, cutting the proxy idle window to 60 ms and ending every session before the first frame.

**Suggested fix:** "Milliseconds of scale silence that end a GATT reading session."

### 6. Minor: withIdleTimeout doc claims it rejects start's promise, which it never touches
`src/ble/types.ts:186`

"Like withTimeout, but the deadline restarts whenever `start`'s callback is invoked: the promise from `start` is rejected only after `ms` of inactivity." The helper rejects its own returned promise and abandons `start`'s, exactly as `withTimeout` does. That distinction is load-bearing here: the mqtt-proxy watcher must call `device?.fireDisconnect()` in its finally (watcher.ts:497 and 627) precisely because the abandoned `waitForRawReading` keeps running with its notification listeners attached. A reader who believes the helper rejects the inner promise would conclude that teardown is redundant and delete it, reintroducing the leak #339 just fixed.

**What goes wrong:** Not a runtime failure; the comment misdescribes the contract that the surrounding teardown depends on.

**Suggested fix:** "...: the returned promise rejects after `ms` with no activity. Like withTimeout, the promise from `start` is abandoned rather than cancelled, so callers must still tear down whatever it holds."

### 7. Minor: The two native call sites change semantics with no test coverage
`src/ble/handler-node-ble/scan.ts:432`

The new tests cover the helper in isolation (tests/ble/types.test.ts) and the mqtt-proxy autonomous path (tests/ble/handler-mqtt-proxy.test.ts). Nothing exercises `handler-node-ble/scan.ts:432` or `handler-noble-shared.ts:602`, and grepping tests/ for `readingTimeoutMs` or "Timed out waiting for a complete scale reading" returns no hits at all, so `ble.session_timeout_sec` (#83) has never had behavioural coverage. These are the only two sites where a user-configurable value feeds the timeout, and this PR changes what that value measures.

**What goes wrong:** A later refactor drops the `onActivity` argument from one of the two native calls (it is the ninth positional parameter of `waitForRawReading`, after three `undefined` placeholders at the proxy sites). That site silently reverts to a fixed deadline, the whole suite stays green, and only a native-handler user with a slow scale notices.

**Suggested fix:** Add one node-ble or noble test with fake timers: a session that receives an adapter-rejected frame just before `readingTimeoutMs` and a complete frame after it must resolve, and a wholly silent session must still reject at `readingTimeoutMs`. That pins both the idle restart and the #83 wiring.

## Worth calling out

- **The regression test pins the exact branch, not just the happy path** (tests/ble/handler-mqtt-proxy.test.ts:1793) "autonomous session survives a weigh-in that starts near the old fixed deadline" idles 55 s, sends a 1-byte frame, asserts `expect(adapter.parseNotification).toHaveLastReturnedWith(null)` so the frame is provably rejected by the adapter, then advances past the old 60 s deadline and asserts no `${PREFIX}/disconnect` was published before a complete frame resolves. Under the previous code the fixed timer fires at 60 s and the finally publishes the disconnect, so the length-0 assertion fails: the test cannot pass without the change, and it specifically pins the subtle half of it, that `onActivity` runs before the adapter gate rather than after.

- **onActivity placed at the only correct point in handleNotification** (src/ble/shared.ts:472) It sits after `if (resolved) return` and before the debug log, the ACK write, and `adapter.parseNotification`. That ordering gives all three properties the change needs at once: frames arriving after the reading completes cannot re-arm a timer the caller is about to clear, adapter-rejected frames still count as liveness, and self-generated ACK and unlock writes do not. Worth noting for the reviewer's grace-timer question: `GraceTimers` (advertisement.ts:102) is fed only from `evaluateAdvertisement` in the scan-results branch and never from the GATT path, and the GATT-side `HoldTimer` is deliberately not restarted by later frames, so neither can fight the new idle timer.

## Questions for you

- The commit cites a real mqtt-proxy session where the user stepped on at T+54 s of a 60 s window. Which ESP32 firmware was that log from? The #296 session guard (GATT_SESSION_MAX_MS, firmware/main.py:45) becomes the only remaining cap on that path after this change, so it matters whether the reporting setup even has it.

- Is the esphome-proxy in scope? src/ble/handler-esphome-proxy/watcher.ts:182 calls waitForRawReading with no bound at all (and scan.ts:155 is only covered by the whole-scan BROADCAST_WAIT_MS), so after this PR that transport is neither deadline- nor idle-bounded. It is pre-existing, but the PR body says all reading call sites were switched and there are six, not four.

- Should ble.session_timeout_sec be exposed in ble-scale-sync-addon/config.yaml and run.sh? It is absent from both today (only force_scale_adapter is plumbed), so HA add-on users cannot reach the option whose meaning this PR changes.

- Confirm the branch plan: this PR carries PR #339's commit 81e418a (the mqtt gatt.ts listener-leak fix and the fireDisconnect teardown). Merge order matters, since the finally blocks reviewed here assume that teardown is present.

alasano:
[comment written by an actual human :)] 

Hey @KristianP26 thanks for taking the time to review all my PRs and sorry for missing some obvious things in some of them, I'll do much broader sweeps for any next ones so you don't have to take as much time on reviews. I'll prepare all the fixes for all the PRs 👌🏻

Awesome repo by the way, thanks a lot for having made it!

KristianP26:
@alasano thanks, you're welcome to contribute

alasano:
The review is addressed in two commits on the branch.

The comments now match the silence-based deadline, including the `POLL_CYCLE_TIMEOUT_MS` budget and the `withIdleTimeout` doc saying it abandons rather than cancels, so callers still tear down. The `session_timeout_sec` schema doc says silence-that-restarts, is scoped to native handlers, and notes node-ble resets the adapter while Noble only disconnects.

The cap is in, nested exactly as you sketched. 90 s on both mqtt-proxy paths, matching `GATT_STALE_MS` so the stale reset's assumption holds, and 3 x the configured silence window on the native paths, which keeps the 900 s poll ceiling out of reach for any `session_timeout_sec` up to 300. A new test feeds an autonomous session adapter-rejected frames every 30 s so the idle timer never fires, and asserts the cap ends it at 90 s with the disconnect published. It fails without the cap.

For the native coverage gap I added two tests that build the same `withIdleTimeout(onActivity => waitForRawReading(..., onActivity))` composition the noble and node-ble sites use. An adapter-rejected frame (parsed to null) just before the deadline restarts it so a later complete frame resolves, and a silent session rejects at the timeout. These pin the composition and the restart, but not the two native handlers end to end. Dropping `onActivity` from the real call sites doesn't fail them, because nothing drives noble or node-ble through a real connect here, and covering that would need a full connect mock none of the existing tests build.

My ESP32 runs the firmware from the 1.24.0 tree, so it has the 180 s session guard from #296. With the host cap at 90 s the host now always gives up first.

On the esphome-proxy question, I left it out on purpose. Its two call sites were unbounded before this PR too, and adding a bound there is a behavior change for hardware I don't have, so I'd keep it separate. My "all reading call sites" wording in the PR body was wrong though, there are six and I switched four.

On the add-on question, it can't reach `session_timeout_sec` today (only `force_scale_adapter` is plumbed), and `qn_protocol_byte` is missing there too. That can be a follow-up PR.

On merge order, this branch carries the listener-leak PR's commit and the teardown you reviewed here comes from it, so reviewing that one first would help. Once it's on dev I'll rebase this and it drops to the single commit.

KristianP26:
`#339` is merged, so this one is clear to rebase too. Same situation as `#350`: the commit it carried is on `dev` as a squash, and `dev` also moved under you with 6b2138e (the legacy-mode notify registrar), so `src/ble/shared.ts` and `tests/ble/shared.test.ts` will both want attention during the rebase.

Worth knowing before you start, since it overlaps your seam directly: 6b2138e routes every notify registration through a registrar that knows whether the session is already closed, so a subscribe that settles after cleanup is unsubscribed on the spot. Your `onActivity` hook and the absolute cap sit above that and should compose without changes, but the conflict hunks in `shared.ts` are where the two meet.

KristianP26 (approved):
Approved and merging. I rebased this onto `dev` myself, since it was waiting on `#339` landing and on 6b2138e moving `src/ble/shared.ts` underneath it.

Two conflicts, both resolved without changing your intent:

- `docs/guide/configuration.md`: the option table had gained two rows on `dev`. Kept those and took your reworded `session_timeout_sec` description, which is the row this PR is actually about.
- `tests/ble/shared.test.ts`: your idle-composition describe block and the notify-teardown block from 6b2138e were appended at the same point. Both kept, both green.

The cap is verified rather than read: raising `GATT_SESSION_ABSOLUTE_MS` to a day fails your test with `expected [] to have a length of 1 but got +0`, so the release of the ESP32 at 90 s is genuinely pinned. Matching it to `GATT_STALE_MS` is the right constant, because it keeps the stale reset's assumption that no session outlives it true rather than approximately true.

Full suite on the rebased head: 133 files, 2270 tests, tsc, eslint and prettier clean.

Thanks for the whole run of these. Seven of your PRs landed today, and the ones that needed a second pass needed it for real reasons that you then closed properly.

PR #352, opened 2026-08-24
fix(docker): install Python 3.12 for garminconnect 0.3.11 and keep the arm/v7 base

Bretos:
## Summary

garminconnect 0.3.4+ requires Python >=3.12, which node:22-slim (Debian
bookworm) ships as 3.11. Switch to the trixie-slim variant (Debian 13,
Python 3.13) for both build and runtime stages.

## Changes

- `garminconnect` updated to `>=0.3.11`
- `base image` updated to `trixie-slim`

## Test Plan

- [x] Tests pass (`npm test`)

## Notes

Tested by running the image manually and used with single scale I own

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,6 +1,6 @@
 # ── Build stage: compile TypeScript ──────────────────────────────────
 ARG BUILDPLATFORM
-FROM --platform=$BUILDPLATFORM node:22-slim AS build
+FROM --platform=$BUILDPLATFORM node:22-bookworm-slim AS build
 
 WORKDIR /app
 
@@ -10,8 +10,18 @@ RUN npm ci --ignore-scripts
 COPY src/ ./src/
 RUN npm run build
 
+# ── Python stage: Python 3.12 for garminconnect 0.3.x ────────────────
+# Debian bookworm (node:22's base) ships Python 3.11; garminconnect 0.3.x
+# requires >=3.12. Switching the whole runtime base to node:22-trixie-slim
+# fixed that but broke linux/arm/v7: Docker's node image has no published
+# arm/v7 manifest for trixie tags (Raspberry Pi Zero 2W / other 32-bit ARM
+# boards). python:3.12-slim-bookworm is a self-contained Python 3.12 built
+# against bookworm's glibc *and* does publish arm/v7, so copy just the
+# interpreter across instead of changing the base OS.
+FROM python:3.12-slim-bookworm AS python
+
 # ── Runtime stage ────────────────────────────────────────────────────
-FROM node:22-slim
+FROM node:22-bookworm-slim
 
 # OCI labels
 ARG VERSION=local
@@ -31,33 +41,49 @@ LABEL org.opencontainers.image.title="BLE Scale Sync" \
 ENV APP_BUILD_CHANNEL=${VERSION}
 ENV APP_BUILD_REF=${VCS_REF}
 
-# System dependencies: BLE (BlueZ + D-Bus), Python (Garmin upload), tini (PID 1),
+# System dependencies: BLE (BlueZ + D-Bus), tini (PID 1),
 # build-essential (node-gyp needs gcc/g++/make for native BLE modules),
-# python3-dev + libffi-dev + libssl-dev + libcurl4-openssl-dev
+# libffi-dev + libssl-dev + libcurl4-openssl-dev
 # (cffi/cryptography/curl_cffi build from source on architectures without
 # pre-built wheels, e.g. linux/arm/v7. curl_cffi is a transitive dep of
-# garminconnect 0.3.x and has no armv7 wheel on PyPI.)
+# garminconnect 0.3.x and has no armv7 wheel on PyPI.), and
+# libsqlite3-0 + libreadline8 + libncursesw6: the copied Python's
+# lib-dynload brings sqlite3/readline/curses extension modules across, but
+# not the system libraries they dlopen — node:22-bookworm-slim doesn't ship
+# them, so any dependency that reaches for one (e.g. an sqlite3-backed
+# cache) would ImportError at runtime with no build-time signal.
 RUN apt-get update && apt-get install -y --no-install-recommends \
       bluez \
       libbluetooth-dev \
       libusb-1.0-0-dev \
       libdbus-1-dev \
       build-essential \
-      python3 \
-      python3-dev \
-      python3-pip \
-      python3-venv \
       libffi-dev \
       libssl-dev \
       libcurl4-openssl-dev \
+      libsqlite3-0 \
+      libreadline8 \
+      libncursesw6 \
       tini \
     && rm -rf /var/lib/apt/lists/*
 
+# Python 3.12 (Garmin upload), copied from the python stage above — see the
+# comment there for why this isn't just `apt-get install python3`.
+COPY --from=python /usr/local/bin/python3.12 /usr/local/bin/python3.12
+COPY --from=python /usr/local/lib/python3.12 /usr/local/lib/python3.12
+COPY --from=python /usr/local/lib/libpython3.12.so* /usr/local/lib/
+COPY --from=python /usr/local/include/python3.12 /usr/local/include/python3.12
+RUN ln -s /usr/local/bin/python3.12 /usr/local/bin/python3 && \
+    ln -s /usr/local/bin/python3.12 /usr/local/bin/python && \
+    ldconfig
+
 WORKDIR /app
 
-# Python dependencies (Garmin upload)
+# Python dependencies (Garmin upload). python3 -m pip, not pip3: the
+# console-script shims weren't copied from the python stage, only the
+# interpreter and site-packages (which already has pip preinstalled).
 COPY requirements.txt ./
-RUN pip3 install --break-system-packages --no-cache-dir -r requirements.txt
+RUN python3 -m pip install --no-cache-dir -r requirements.txt
 
 # Node.js dependencies (production only)
 COPY package.json package-lock.json ./
--- a/docker-entrypoint.sh
+++ b/docker-entrypoint.sh
@@ -6,7 +6,7 @@ CMD="${1:-start}"
 reset_bt_adapter() {
   if command -v btmgmt >/dev/null 2>&1; then
     echo "Resetting Bluetooth adapter..."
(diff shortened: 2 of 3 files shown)


KristianP26:
Retargeted the base branch from `main` to `dev`. Contributions land on `dev` and reach `main` only through a release promotion, since `main` is what release-please cuts tags from. Nothing about your commits changes, only where they merge.

A full review of this one is still outstanding, and I want to give the Python-interpreter copy the attention it deserves rather than wave it through: copying `python3.12`, its stdlib and `libpython3.12.so` out of `python:3.12-slim-bookworm` into `node:22-bookworm-slim` leaves the C extension modules in `lib-dynload` depending on system libraries that the Node slim image does not ship (`libexpat`, `libsqlite3`, `libbz2`, `liblzma`, `libncursesw`, `libreadline`, `libuuid`, `libgdbm`). Some of those only bite at import time inside a dependency, so the image can build clean and still fail on the first Garmin upload.

Two things that would settle it quickly:

1. The output of `docker run --rm <your image> python3 -c "import ssl, sqlite3, hashlib, lzma, bz2, ctypes, curl_cffi, garminconnect; print('ok')"` for a linux/arm/v7 build as well as amd64. That import list is what the Garmin bridge actually pulls in.
2. Confirmation that a real weigh-in uploaded to Garmin from the built image, since `garminconnect` 0.3.2 to 0.3.11 spans the auth-flow rework and the bridge does its own token handling.

Thanks for chasing the arm/v7 manifest gap rather than just switching the base and moving on. That part of the diff is exactly right.

Bretos:
Hey,

thank you for your quick reply!

1. import list for `amd64`: 
```
docker run --rm new_ble-scale-sync:latest python3 -c "import ssl, sqlite3, hashlib, lzma, bz2, ctypes, curl_cffi, garminconnect; print('ok')"
ok
```

At this time I do not own any `arm/v7` to test that on :(

2. garmin upload is the exact reason why I prepared this change - it stopped working for me when using older version of `python-garminconnect`. Currently it works as expected:

```
[Garmin] Loading tokens from /home/node/.garmin_tokens
[Garmin] Authenticated.
[Garmin] Uploading body composition...
[Garmin] Upload successful!
2026-08-24 08:15:01.413 [Garmin] Garmin upload succeeded.
2026-08-24 08:15:01.413 [Sync] Done.
```

KristianP26 (changes requested):
Follow-up to my earlier comment: I have now built and exercised this, so here is the full review.

The diagnosis is right and the fix is the right shape. I reproduced your image layout locally (`python:3.12-slim-bookworm` stage, interpreter and stdlib copied into `node:22-bookworm-slim`, the same apt list, `python3 -m pip install -r requirements.txt`) and everything the Garmin bridge touches works. Two things need changing before merge, neither of them in the Python approach itself.

## What I verified

**The arm/v7 reasoning is exactly right, and I confirmed it against the registry** rather than taking it on trust:

    node:22-trixie-slim        amd64, arm64/v8, ppc64le, s390x          (no arm/v7)
    node:22-bookworm-slim      amd64, arm/v7, arm64/v8, ppc64le
    python:3.12-slim-bookworm  amd64, arm/v7, arm64/v8, 386, ppc64le

`.github/workflows/docker.yml:62` builds `linux/amd64,linux/arm64,linux/arm/v7`, so a straight switch to trixie would have broken the Pi Zero 2W and every other 32-bit ARM board on the next release. Catching that and solving it with a targeted interpreter copy instead of reverting is the best part of this PR.

**The runtime prerequisites hold.** `/usr/local/lib` is already in the default `/etc/ld.so.conf` on bookworm, so `libpython3.12.so.1.0` resolves (`ldd /usr/local/bin/python3.12` confirms it). `python3 -m pip` resolves from the copied `site-packages` with no console-script shims. `timeout` is present, GNU coreutils 9.1.

**Source builds still work, which is what arm/v7 actually needs.** Dropping `python3-dev` worried me, since `curl_cffi` has no arm/v7 wheel and compiles `cffi` from source there. It is fine: `/usr/local/include/python3.12/Python.h` and `/usr/local/lib/python3.12/config-3.12-*-linux-gnu` both come across in the copy, and I proved it by forcing a from-source rebuild in the image:

    python3 -m pip install --no-binary :all: --force-reinstall cffi   ->   cffi 2.1.1 built and imports

**garminconnect 0.3.11 is API-compatible with the bridge.** I introspected the installed package against what `garmin-scripts/` actually calls:

    add_body_composition(self, timestamp, weight, percent_fat=None, percent_hydration=None,
                         visceral_fat_mass=None, bone_mass=None, muscle_mass=None, basal_met=None,
                         active_met=None, physique_rating=None, metabolic_age=None,
                         visceral_fat_rating=None, bmi=None)
    login(self, tokenstore=None) -> tuple[str | None, str | None]
    resume_login present: True

Every keyword `garmin_upload.py:78` passes exists, `login(token_dir)` and `resume_login(state, code)` match `setup_garmin.py`, and both scripts run under the new interpreter. So the 0.3.2 to 0.3.11 span carries no breaking change for us.

## Findings

### 1. Major: the interpreter copy silently breaks three stdlib modules
`Dockerfile:64`

The copy brings `lib-dynload` across but not the system libraries those extension modules link against, and `node:22-bookworm-slim` does not ship them. Measured in the built image:

    OK    ssl, hashlib, lzma, bz2, ctypes, zlib, uuid, decimal, _gdbm, pyexpat, xml.etree
    FAIL  sqlite3   ImportError: libsqlite3.so.0: cannot open shared object file
    FAIL  readline  ImportError: libreadline.so.8: cannot open shared object file
    FAIL  curses    ImportError: libncursesw.so.6: cannot open shared object file

I checked all Python in the repository (`garmin-scripts/garmin_upload.py`, `garmin-scripts/setup_garmin.py`, `ble-scale-sync-addon/merge_last_weights.py`) and none of the three is imported, directly or through `garminconnect`, `curl_cffi`, `requests`, `yaml` or `dotenv`. All of those import cleanly. So nothing is broken today.

**What goes wrong:** a later `garminconnect` picks up a dependency with an sqlite3-backed cache (`requests-cache` is the obvious candidate). The wheel installs without complaint at build time, CI stays green because CI never runs the Garmin bridge, and the failure surfaces as an `ImportError` on a user's first weigh-in after the upgrade, on the exporter that is hardest for them to debug. The Docker layer where it broke is months old by then.

**Suggested fix:** add the three runtime libraries to the existing apt block. They are small and it removes the whole class of problem:

    libsqlite3-0 \
    libreadline8 \
    libncursesw6 \

If you would rather keep the image lean, that is a defensible call, but then say so in the comment above the `COPY --from=python` lines: note that `lib-dynload` modules needing libraries absent from the Node base will fail at import, and name the three. A future reader needs to know the copy is deliberately partial.

### 2. Major: `deps(docker)` is not a configured commit type, so this ships invisibly and cuts no release
`.release-please-config.json:13`

`changelog-sections` lists `feat`, `fix`, `perf`, `revert`, `docs`, `refactor`, `chore`, `ci`, `test`, `build`, `style`. `deps` is not among them, and release-please ignores commit types it does not know: no version bump, no changelog entry.

**What goes wrong:** this merges, the release PR either does not appear or appears without it, and users still on Python 3.11 keep hitting the `garminconnect` install failure with nothing in the changelog to tell them a fix exists. The one commit in the tree that repairs Garmin uploads is the one commit release-please cannot see.

**Suggested fix:** retitle to `fix(docker): install Python 3.12 for garminconnect 0.3.11 and pin the bookworm base`. That lands it under "Fixed" and cuts a patch release, which is what this actually is: a user-visible breakage repair, not a routine dependency bump.

### 3. Minor: the btmgmt timeout is a real improvement in the wrong PR
`docker-entrypoint.sh:9`

Wrapping the two `btmgmt` calls in `timeout 5` is correct, and I confirmed `timeout` exists in the new base. But it has nothing to do with the Python bump, and with the retitle above it would land in the changelog under a heading that does not describe it.

**Suggested fix:** split it into its own commit on the same branch, `fix(docker): bound the btmgmt adapter reset so a wedged controller cannot stall startup`. Two commits, one PR, is fine here since the repo squashes only single-purpose PRs.

### 4. Minor: no upper bound on garminconnect
`requirements.txt:1`

`garminconnect>=0.3.11` will happily pull a future 0.4 that reworks the auth flow again, exactly as 0.3 did to 0.2, and the first sign will be a user's failed upload rather than a failed build.

**Suggested fix:** `garminconnect>=0.3.11,<0.4`. The bridge already carries explicit pre-0.3 token migration logic in both scripts, so it is clearly a library whose majors matter.

## One thing to confirm

I could only build and run amd64 here. Everything above (the loader path, the missing three libraries, the from-source `cffi` build) is architecture-independent, and the manifest check proves arm/v7 exists for both images, but the arm/v7 build itself compiles `curl_cffi` from source and that is the slow, fragile path. If you have a way to run

    docker buildx build --platform linux/arm/v7 .

against your branch, the build log alone would settle it. If not, say so and I will run it before merging rather than leave it to the release.

Thanks for this one. The trixie regression would have been very easy to miss until a Pi user reported it.

KristianP26 (approved):
Approving. I settled the arm/v7 question locally rather than leaving it with you, since you do not have the hardware.

I built this branch for `linux/arm/v7` under QEMU and ran the import list against the resulting image:

```
docker buildx build --platform linux/arm/v7 -t pr352-armv7:test --load .
docker run --rm --platform linux/arm/v7 pr352-armv7:test python3 -c "
import ssl, sqlite3, hashlib, lzma, bz2, ctypes, uuid, readline, curses
from xml.parsers import expat
import xml.etree.ElementTree as ET
import curl_cffi, garminconnect, yaml, dotenv, zlib, decimal, socket
print('IMPORTS OK')"
```

`IMPORTS OK`, on `armv7l` with Python 3.12.14. That covers every module I was worried about, including the three you added libraries for and the ones I listed that you did not (`expat`, `uuid`, `lzma`, `bz2`), so the copied `lib-dynload` is fully satisfied on 32-bit ARM and not only on amd64. I also introspected the installed package in that image: `add_body_composition` and `login` both match what the bridge calls.

The `timeout 5` around `btmgmt` is a good catch and unrelated to the rest, but it is small and in the same file, so I am taking it here rather than asking you to split it.

One change from my side before merging: the title moves from `deps(docker):` to `fix(docker):`. The squash commit takes the PR title, and `deps` is not one of the changelog sections in the release config, so as written this would have shipped without a changelog entry. Your commits on the branch were already `fix(docker):`; only the title was out of step.

Issue #280, opened 2026-07-16
node-ble: ManufacturerData is never plumbed into BleDeviceInfo, silently degrading every manufacturer-id matcher on Linux


KristianP26:
## Problem

`BleDeviceInfo.manufacturerData` is never populated on the node-ble (Linux/BlueZ) transport, which is the project's primary path and the one the Raspberry Pi production install uses. BlueZ exposes it as the `ManufacturerData` D-Bus property on `org.bluez.Device1`, but we never read it.

Every adapter that fingerprints on manufacturer data is therefore silently degraded on Linux native:

- `beurer-bf720.ts` - Beurer company id `0x0611` no-name path
- `mi-scale-2.ts` - uses `0x0611` to *exclude* Beurer scales
- `qn-scale.ts`, `eufy-p2.ts` - vendor id checks
- `lefu-signature.ts` / `hutbit.ts` - the Lefu OEM signature (#278)

None of them fail loudly. They just quietly take a different branch.

## How it surfaced

#278: a Hutbit 218008 with Lefu OEM stock branding advertises as `SWAN` and is claimed by its `0x02AC` manufacturer signature. That works on esphome-proxy, mqtt-proxy scan, and both noble paths (target-MAC was fixed in 9d17405), but on node-ble the signature is invisible, so the unit still falls to MGB, which cannot parse its 8-byte frames.

## Where

- `src/ble/handler-node-ble/discovery.ts:209`
- `src/ble/handler-node-ble/scan.ts:271-275`

Both build a `BleDeviceInfo` without `manufacturerData`.

## Fix sketch

Read the `ManufacturerData` property (a D-Bus `a{qv}`, company id to byte array) and map it to the existing `{ id, data }` shape, the same normalization `parseMfgData` does for noble in `handler-noble-shared.ts:55`. Adapters need no change.

## Test

A node-ble discovery fixture carrying manufacturer data resolves to the signature-gated adapter, mirroring `tests/scales/hutbit.test.ts`.

KristianP26:
Fixed and closed. Every site on the node-ble transport that builds a device record now carries the advertisement, so the manufacturer-id matchers work on Linux.

The two sites this issue named:

- `src/ble/handler-node-ble/scan.ts`, the MAC-pinned pre-connect match. It was throwing away an advertisement snapshot it had already read, which meant the passive branch and the bond-on-timeout retry both keyed on the name alone.
[...]

PR #353, opened 2026-08-24
feat(scales): add Xiaomi Mi Smart Scale 2 XMTZC04HM adapter

O-Palda:
## Summary

- Add a dedicated broadcast-only adapter for the Xiaomi Mi Smart Scale 2 XMTZC04HM / `MI SCALE2`.
- Decode its 10-byte `0x181D` Weight Scale Service frame with openScale's flag mapping: bit 0 = lb, bit 4 = jin/catty, kg when neither is set, bit 5 = stable, bit 7 = removed.
- Accept a stable weight immediately, so the reading skips the 12 s impedance grace timer and goes through the normal dedup window.
- Match the exact `MI SCALE2` name on the node-ble pre-connect and auto-discovery paths, which see no service or manufacturer data; nameless advertisements still require Huami company ID `0x0157` plus the exact 10-byte `0x181D` frame.
- Keep MIBCS/MIBFS parsing and body-composition behaviour unchanged.
- Document the XMTZC04HM as weight-only with a BMI/profile estimate, and correct the adapter counts.

## Verification

kg and pounds are both confirmed against the hardware over an ESPHome proxy. jin/catty follows the openScale / ble_monitor / xiaomi-ble mapping and is covered by a synthetic fixture only — Zepp Life offers this model kg and lb, with no jin option, so a real jin frame cannot be captured.

## Validation

- `npx vitest run tests/scales/mi-scale-2.test.ts tests/scales/xiaomi-mi-scale-legacy.test.ts tests/scales/registry-collision.test.ts`
- `npm run lint`
- `npm run format:check`
- `npm run build`
- `npm test`


KristianP26 (changes requested):
The PR teaches MiScale2Adapter to decode the 10-byte Weight Scale Service (0x181D) advertisement of the Xiaomi Mi Smart Scale 2 XMTZC04HM, gating the nameless match on the Huami company id 0x0157, and short-circuits computeMetrics to a profile estimate when impedance is 0. The Huami gate is well designed and provably collision-free, the length-plus-UUID discriminator matches what ESPHome uses, and the shared normalizeServiceUuid extraction is a clean cleanup. However the unit-flag decode is wrong: the code reads bit1 as "kg", but bit1 is the SIG Time Stamp Present flag and is set on every 10-byte frame, so the kg branch is taken unconditionally and a scale configured in pounds reports a weight 10.23 percent too heavy. Two further issues need attention before merge: the computeMetrics early return silently rewrites body composition for existing MIBCS/MIBFS weight-only readings (muscleMass nearly doubles), and isComplete can never be true for this variant, which forces a 12 second grace delay per reading and lets the two continuous-mode watchers enqueue duplicate exports because the grace path bypasses the dedup window.

## Findings

### 1. Blocker: Unit flag bits are wrong: bit1 is Time Stamp Present, not kg, so lb readings are 10.23% too heavy
`src/scales/mi-scale-2.ts:176`

The 10-byte 0x181D payload is the Bluetooth SIG Weight Scale Measurement layout: 1 flags byte + 2 weight bytes + 7 timestamp bytes. SIG flags bit0 is the unit (0 = SI kg, 1 = Imperial lb) and bit1 is Time Stamp Present. The PR's own capture proves the layout: `22 6e46 b207 01 14 11 2f 33` decodes its tail as year 0x07b2 = 1970, month 01, day 0x14 = 20, hour 0x11 = 17, minute 0x2f = 47, second 0x33 = 51. Because the frame is 10 bytes precisely BECAUSE the timestamp is present, bit1 (0x02) is set on every well-formed frame. The code tests `flags & 0x02` first and calls it kg, so the kg branch is taken unconditionally, and the `flags & 0x04` lb branch, the `flags & 0x10` jin branch and the trailing `return null` are all dead code. 0x04 is not a Xiaomi unit bit at all (SIG: User ID Present).

Authoritative mapping, from openScale's own MiScaleHandler.kt for the v1 / 10-byte / 0x181D frame: LBS = `status and 0x01`, catty/jin = `status and 0x10`, stabilized = `status and 0x20`, removed = `status and 0x80`, and kg is the absence of both unit bits; weight = `if (isLbs || isCatty) raw / 100 else raw / 200`. That is exactly what the 13-byte v2 parser two functions above in this same file already implements (`isLbs = (c0 & 0x01) !== 0`, `isCatty = (c1 & 0x40) !== 0`). ESPHome's xiaomi_miscale component corroborates the frame gate (`uuid == 0x181D && size == 10`) and lists the lbs control byte as 0x03 / 0xb3, i.e. bit0 set, never bit2.

The single capture in the PR cannot distinguish the two mappings: I ran flags 0x22 through both readings and both yield 90.15 kg. Neither can the test suite, because there is no lb or jin case on the 0x181D path.

Only jin escapes, by luck: `raw / 200` (the branch actually taken) equals openScale's `(raw / 100) * 0.5`.

**What goes wrong:** A user sets the XMTZC04HM to pounds in Mi Fit and stands on it at 180.00 lb (81.65 kg). The scale emits a stable frame with flags 0x23 (bit0 lb | bit1 timestamp | bit5 stable) and raw weight 18000 hundredths of a pound. I ran this through the patched adapter: it returns `{ weight: 90, impedance: 0 }`. The reading is 8.35 kg (18.4 lb) too heavy, a constant +10.23% error (0.005 / 0.00453592 = 1.1023), and it is exported as a genuine weigh-in to Garmin, InfluxDB, MQTT and every other configured exporter. On firmware that emits the ESPHome-observed 0x03 for lb (bit5 clear) the reading is dropped instead, so an lb user gets either silence or a 10% wrong number, never a correct one.

**Suggested fix:** Mirror the v2 parser in this same file:

    const isLbs = (flags & 0x01) !== 0;
    const isCatty = (flags & 0x10) !== 0;
    let weight: number;
    if (isLbs) {
      weight = (rawWeight / 100) * 0.453592;
    } else if (isCatty) {
      weight = (rawWeight / 100) * 0.5;
    } else {
      weight = rawWeight / 200;
    }

Drop the `return null` fallthrough (kg is the default, not an error case) and correct the comment on line 165 to `bit 7 = weight removed, bit 5 = stable, bit 4 = jin, bit 1 = timestamp present, bit 0 = lb`. Add parseServiceData tests for a stable lb frame (0x23) and a stable jin frame (0x32) so the mapping is pinned, since the existing kg-only test passes under either reading.

### 2. Major: computeMetrics early return silently rewrites body composition for existing MIBCS/MIBFS weight-only readings
`src/scales/mi-scale-2.ts:197`

The `if (impedance <= 0) return buildPayload(weight, 0, {}, profile)` guard is not reachable only from the new XMTZC04HM path, despite what the comment above it says. The existing 13-byte v2 frame also yields impedance 0 whenever the impedance-present bit is clear (parseFrame lines 126 to 129), and every transport publishes that weight-only reading once the grace window expires: node-ble handler-node-ble/broadcast.ts:140, noble handler-noble-shared.ts:371, esphome handler-esphome-proxy/scan.ts:83 and watcher.ts:50, mqtt handler-mqtt-proxy/watcher.ts:93, plus the single-shot `weightOnlyFallback` return at handler-mqtt-proxy/scan.ts:219. Each of those RawReadings reaches `adapter.computeMetrics` (runtime/processor.ts:149 and the four scanAndRead sites).

I measured the delta by running both code paths. Male, 35y, 178 cm, 82.0 kg, impedance 0:
  before: bodyFat 21.38, water 53.93, bone 3.29, muscleMass 31.81, visceralFat 15
  after:  bodyFat 22.91, water 56.28, bone 2.66, muscleMass 60.56, visceralFat 11
Female, 42y, 165 cm, 68.0 kg, impedance 0:
  before: bodyFat 31.48, water 48.92, bone 2.76, muscleMass 20.16, visceralFat 1
  after:  bodyFat 34.23, water 48.01, bone 1.88, muscleMass 42.84, visceralFat 18

The new numbers are arguably the project-correct ones (the old path fed MiScaleCalc.muscle()'s skeletal percentage into muscleMass, which is the #253 bug fixed elsewhere), so I am not asking for a revert. The problem is that the change is invisible: the PR title, body, code comment and docs row all attribute it to the XMTZC04HM only.

**What goes wrong:** An existing Mi Scale 2 (MIBFS) user in continuous mode steps off before the impedance frame arrives. The 12 second grace timer publishes the weight-only reading. Before this PR their muscleMass exported as 31.81 kg and visceralFat as 15; after it, 60.56 kg and 11 from the identical frame. In the female example visceralFat jumps from 1 to 18. Their Garmin and InfluxDB series show a step discontinuity on the day the update lands, with no changelog entry explaining it.

**Suggested fix:** Keep the early return, but stop describing it as XMTZC04HM-specific. Reword the comment to say it applies to any zero-impedance Mi frame (v1 always, v2 when the impedance bit is clear), state the behaviour change in the commit body so release-please surfaces it, and add a line to the supported-scales entry noting that a Mi Scale 2 weigh-in with no impedance frame now uses the shared profile estimate. A dedicated test asserting the v2 weight-only path also routes through buildPayload would document the intent.

### 3. Major: isComplete can never be true for the XMTZC04HM: 12 s delay per reading, and watchers enqueue un-deduped duplicates
`src/scales/mi-scale-2.ts:189`

`isComplete` still requires `impedance > 0`, but the 0x181D frame always returns `impedance: 0`, and `preferPassive` is true. So `evaluateAdvertisement` (src/ble/advertisement.ts:75 to 80) classifies every XMTZC04HM advertisement as `partial`, forever.

I traced all four transports and confirmed a reading is eventually produced everywhere, so Suspicion B is not a total-failure bug: node-ble broadcast.ts:139-148 has its own grace timer, handler-noble-shared.ts:371 uses GraceTimers, esphome scan.ts:83 and watcher.ts:50 use GraceTimers, mqtt watcher.ts:93 uses GraceTimers, and the mqtt single-shot scan falls back at scan.ts:219 with no wait. But on every streaming transport the user waits the full IMPEDANCE_GRACE_MS (12 s, src/ble/types.ts:105) for a value the adapter already knows is final.

The secondary consequence is worse. In both continuous-mode watchers the grace callback pushes straight onto the queue and deliberately skips the dedup window: handler-esphome-proxy/watcher.ts:50-56 (`this.queue.push(gr)`) versus `pushDeduped` at :138, and handler-mqtt-proxy/watcher.ts:93-100 versus the `dedup.shouldEmit` guard at :235. GraceTimers.hold re-arms after the timer fires, because the callback deletes the timer entry before invoking (src/ble/advertisement.ts:119 to 131). Previously this mattered little, since a MIBFS eventually sends an impedance frame that cancels the grace and routes through the dedup window; for the XMTZC04HM the un-deduped grace path is the only path.

The in-repo precedent is src/scales/xiaomi-s800.ts:177-180, the other preferPassive weight-only adapter, which overrides `isComplete` to `reading.weight > WEIGHT_MIN` for exactly this reason (its comment: "Broadcast weight has impedance 0; accept any plausible weight").

**What goes wrong:** Continuous mode over an ESPHome proxy. A user steps on the XMTZC04HM and stands still while it settles. At t=0 the stable 0x22 frame arrives, classified partial, grace armed. At t=12 s the grace fires, the reading is pushed directly onto the queue bypassing DedupWindow, and 90.15 kg is exported. The scale is still advertising stable frames at t=12.5 s, which re-arms the grace, and at t=24.5 s a second identical 90.15 kg weigh-in is exported. Garmin Connect and InfluxDB record two weigh-ins for one step-on.

**Suggested fix:** Preferred: split the XMTZC04HM into its own adapter class, as the repo already does for renpho / renpho-msc04 / renpho-es26bb, with `preferPassive = true`, `isComplete(r) { return r.weight > 10; }` and its own unique priority. That resolves the frame immediately through the dedup window, removes the need for both the shared parseServiceData branch and the computeMetrics early return, and leaves MIBCS/MIBFS behaviour untouched. Remember to bump the adapter count (29 to 30 protocol adapters in docs/guide/supported-scales.md line 15 and the logo `adapters.exporters` figure) if you take this route. If you prefer to keep one class, at minimum route the grace-elapsed push through `pushDeduped` in both watchers.

### 4. Minor: The nameless Huami 0x181D branch is unreachable on both native transports
`src/scales/mi-scale-2.ts:90`

The new branch needs `manufacturerData` and `serviceData` present in the same BleDeviceInfo. Only the two proxy transports supply both on the matching path: src/ble/handler-esphome-proxy/advert.ts toBleDeviceInfo, and src/ble/handler-mqtt-proxy/scan.ts:31-49. On the native transports it cannot fire:

- noble: every resolveAdapter call site omits serviceData (handler-noble-shared.ts:312, :486, :578, :653), so a nameless XMTZC04HM is never claimed there.
- node-ble MAC-pinned: the pre-connect match at handler-node-ble/scan.ts:279 passes `{ localName: name, serviceUuids: [] }` and nothing else, which the comment at :272-278 already documents as a known limitation.
- node-ble auto-discovery: handler-node-ble/discovery.ts:206 skips nameless devices outright (`if (!name) continue`).
- node-ble post-discovery: scan.ts:327-338 DOES carry both fields via `...advert` (the #280 fix), but the new branch sits after the `chars && chars.length > 0` early return at lines 82 to 85, which returns false because an XMTZC04HM has no Mi vendor history characteristic. The device then falls through to Standard GATT.

In practice the XMTZC04HM advertises as MI_SCALE and is caught by KNOWN_PREFIXES, so this is a robustness gap rather than a live failure. The new test asserting the nameless case passes, but it only exercises the proxy shape, so it reads as broader coverage than actually exists.

**What goes wrong:** A nameless (or name-suppressed) XMTZC04HM is MAC-pinned on a Raspberry Pi using the node-ble handler. The pre-connect match at scan.ts:279 sees only an empty name, so no adapter is pre-matched and the code GATT-connects. Post-discovery, mi-scale-2 matches() reaches line 83 with characteristicUuids populated, does not find the Mi vendor history char, and returns false, even though the very same BleDeviceInfo carries the Huami 0x0157 manufacturer data and the 10-byte 0x181D service data the new branch was written for. Standard GATT claims it and subscribes to 0x2A9D, which this broadcast scale does not stream, so the cycle times out.

**Suggested fix:** Move the `hasLegacyWssAdvert` check above the `characteristicUuids` early return at lines 82 to 85. It is a positive vendor fingerprint (company id plus exact service-data length), so hoisting it cannot reintroduce the Beurer BF950 / #255 regression that the chars early return exists to prevent. Optionally add `manufacturerId: 0x0157` and `'181d'` to the MatchDescriptor at line 39 so the declared claims stay a superset of what matches() can claim (registry-check skips overlap analysis for `custom: true`, so this will not fail CI either way).

### 5. Minor: Docs row drops the body-composition note and misses a Known Limitations entry; table column no longer aligns
`docs/guide/supported-scales.md:18`

Three separate points on the one changed row.

1. The trailing "Uses scale's own body comp values" was removed and nothing replaced it. Removing it is defensible (it was inaccurate: this adapter computes body composition locally via the MiScaleLib port, it does not read vendor values), but the row now says nothing about how body composition is produced, and the XMTZC04HM is a weight-only scale whose composition is a BMI/profile estimate. Every other weight-only entry in this table spells that out and has a Known Limitations row: Renpho R-MSC04, Eufy Smart Scale P2/P2 Pro, Renpho ES-CS20M broadcast-only variants.

2. Prose accuracy is otherwise good: the 0x181B-with-impedance versus 0x181D-weight-only split is correct, and "Huami manufacturer-ID-tagged" accurately describes the new gate. No em dash and no double dash were introduced anywhere in the diff (I checked every added line; the only non-ASCII character is a rightwards arrow at the lb-to-kg comment, which the project uses elsewhere).

3. The new first-column cell is 70 characters wide while every other row in that table is 71, so the pipes no longer line up in the raw file. CI will not catch it: `npm run format:check` covers only `src/**/*.ts` and `tests/**/*.ts`, and the file is already not Prettier-clean on the base branch, so this is a pre-existing style gap the PR widens by one row.

The "29 protocol adapters" count on line 15 stays correct, since no adapter was added.

**What goes wrong:** A prospective XMTZC04HM owner reads the supported-scales table, sees the Mi Scale 2 row alongside MIBFS with no caveat, and expects impedance-measured body fat. They set it up, get BMI-estimated figures indistinguishable in shape from real ones, and file an issue about wrong body-fat percentages, exactly the ticket flow the R-MSC04 and Eufy P2 Known Limitations rows were written to prevent.

**Suggested fix:** Append to the cell: "XMTZC04HM is weight-only, so body composition is estimated from BMI, see Known Limitations below." Add a matching Known Limitations row: "**Xiaomi** Mi Scale 2 (XMTZC04HM) | Weight only. The 0x181D advertisement carries no impedance, so body composition is estimated from BMI (Deurenberg formula)." Pad the first-column cell to 71 characters so the table lines up with its neighbours.

### 6. Minor: The frame's 7-byte timestamp is discarded with no comment on why
`src/scales/mi-scale-2.ts:172`

Bytes [3..9] of the 10-byte frame are the SIG timestamp (the PR's own capture decodes to 1970-01-20 17:47:51), and the parser reads only [0] and [1..2]. `ScaleReading.timestamp` exists specifically for frames whose protocol carries an age, and consumers route timestamped readings into the cache-replay pipeline (src/interfaces/scale-adapter.ts:27-34). openScale itself describes the 10-byte 0x181D payload as a history record format. v1.24.0 shipped a fix for exactly this class of mistake on the QN 0xB4 record, where a stored history entry was published as a live weigh-in. Nothing in the diff says whether the timestamp was considered.

**What goes wrong:** If a unit ever replays a stored 0x181D record in its advertisement (the format carries a timestamp precisely because it can), the adapter publishes it as a live weigh-in with no staleness check, repeating the v1.24.0 stale-record incident. Even absent that, a reviewer six months from now cannot tell whether the timestamp was examined and rejected or simply overlooked.

**Suggested fix:** Add one comment line recording the decision, for example: "Bytes [3..9] are the SIG timestamp. It is not used as ScaleReading.timestamp because many units ship with the clock unset (the reference capture reads 1970), which would make a staleness guard unusable." If a reporter capture ever shows a non-1970 clock, revisit surfacing it.

## Worth calling out

- **The Huami company-id gate is genuinely collision-free, and I verified it against the whole registry** (src/scales/mi-scale-2.ts:91) Refusing to claim bare 0x181D and requiring company id 0x0157 plus an exactly-10-byte 0x181D service-data entry is the right call, and it holds up under adversarial checking. Every adapter that outranks Mi Scale 2 (priority 210) bows out for this device: Beurer BF720 at 220 requires company id 0x0611 for its nameless fallback (beurer-bf720.ts:287), Renpho ES-26BB at 230 and R-MSC04 at 235 are name-exact, Renpho ES-WBE28 at 240 requires 'renpho' in the name (renpho.ts:130), QN Scale at 250 requires a QN vendor service or the FFE1+FFE3 characteristic pair for nameless devices (qn-scale.ts:625), Senssun at 260 is name-exact, and Eufy P2 at 270 requires company id 0xFF48. The existing collision fixture `{ localName: 'GenericScale', serviceUuids: ['181d'] }` at tests/scales/registry-collision.test.ts:77 still resolves to Standard GATT, and I confirmed the four registry and resolution test files stay green with the patch applied. The UUID-plus-length discriminator also matches ESPHome's own `parse_header_` gate for this frame, and factoring the duplicated UUID expansion into `normalizeServiceUuid` is a clean improvement over the inline block it replaces.

## Questions for you

- A raw DEBUG `Advert:` line from an XMTZC04HM switched to pounds (and ideally to jin), so the flag byte can be confirmed against openScale's bit mapping rather than inferred. The single kg capture in the PR yields 90.15 kg under both the current code and the openScale mapping, so it cannot settle the question.

- Confirmation that the reporter's unit really does emit the Huami 0x0157 manufacturer-data section in the SAME advertisement PDU as the 0x181D service data. The test fixture's `70879eede5e7` looks like a reversed MAC address; a real capture would confirm the gate is not dead in the field.

- Whether the unit is actually nameless on the reporter's transport, or whether it advertises MI_SCALE and is already caught by KNOWN_PREFIXES. This decides how much the finding about the unreachable native-transport branch matters.

- How long the scale keeps emitting stable (bit 5 set) frames after the weight settles, and whether it switches to the removed flag (bit 7) promptly on step-off. This sizes the duplicate-export risk from the 12 second grace timer re-arming.

O-Palda:
Thanks — you were right on the flag mapping, and I have hardware confirmation now. Revised from ESPHome captures of an XMTZC04HM switched to pounds.

### The unit flags

The captured stable pounds frame is `238e4eea070818103913`: flags `0x23`, raw `0x4E8E` = 20110 hundredths of a pound = 201.10 lb = 91.217 kg, matching the scale's own display. Bit 0 is clear on every frame of the kg session and set on every frame of the pounds session; bit 2 — the bit my first patch used for pounds — never appears. **Under the patch as originally submitted this frame decoded as 100.55 kg.**

kg and pounds are both confirmed on hardware. jin/catty is implemented from the openScale / ble_monitor / xiaomi-ble mapping and is covered by a synthetic fixture only: Zepp Life offers this model kg and lb, with no jin option, so I cannot capture a real jin frame.

### Changes

1. Split the XMTZC04HM into a dedicated passive adapter. `mi-scale-2.ts` retains its original behaviour, including for weight-only (impedance-0) v2 frames, so no existing MIBCS/MIBFS reading changes.
2. Corrected the `0x181D` flag mapping, with tests for kg, pounds, jin/catty, unstable and removed frames.
3. A stable XMTZC04HM reading is complete immediately, so it never enters the 12 s impedance grace path and normal deduplication applies.
4. Added an exact `MI SCALE2` name match for the node-ble pre-connect and auto-discovery paths, which see no service or manufacturer data. Nameless advertisements still require the `0x0157` + 10-byte `0x181D` fingerprint. MIBFS and generic WSS scales remain excluded.
5. Documented the weight-only BMI estimate, added a Known Limitations row, and corrected the adapter-count wording: 30 protocol adapters plus the generic fallback. I left the JSON-LD block at 31, which counts the generic adapter — the prose and the JSON-LD already disagreed by one before this PR, so I did not silently unify them.
6. The seven-byte timestamp is deliberately not surfaced as `ScaleReading.timestamp`: live readings carry it as well, while that field marks a historical cache replay in this runtime. The code records the decision.

### Two notes on the review

- **Finding 4:** the XMTZC04HM does expose the Mi vendor history characteristic. The ESPHome GATT dump lists `00002a2f0000351221180009af100700` at handle 0x22, alongside `1531`/`1532`/`1542`/`1543`, so the post-discovery fall-through to Standard GATT you describe does not happen on this device. The real gap was the name-only pre-connect path, which the exact-name branch now covers.
- **Finding 6:** the clock is not generally unset. The 1970 timestamp in my first capture came from a unit that had never been paired; after one Zepp Life pairing every frame carries the correct wall-clock time (`2026-08-24 16:57:19` in the capture above).

### Possible follow-up

The scale keeps re-advertising its last stable record for minutes — I logged one still being broadcast 2 min 38 s after it was taken — while `DEDUP_WINDOW_MS` is 30 s in both watchers. So a stored record can still be exported more than once even with `isComplete` now immediate. The `0x181D` timestamp is exactly what a staleness guard would need; it would have to tolerate an unset (1970) clock, and the docs would then want a note that pairing once with Zepp Life sets it. Happy to open a separate issue if that is worth doing.

Focused adapter, registry-collision, lint, format, TypeScript build and the full test suite all pass.

O-Palda:
Attaching the full capture behind the numbers above, unedited.

This run is on the released **1.24.0**, not on this branch — you can tell because Standard GATT claims the device at `16:57:20.268` once the local name drops out of the advertisement, which the patch prevents. So it doubles as a record of the current released behaviour on this hardware.

### The pounds weigh-in, frame by frame

| log time | flags | bits set | raw | frame clock | lb → kg | as the first patch decoded it |
| --- | --- | --- | ---: | --- | ---: | ---: |
| 16:57:14.421 | `0xa3` | 0,1,5,7 | 20110 | 16:55:43 | 91.22 | 100.55 |
| 16:57:15.764 | `0x03` | 0,1 | 8340 | 16:57:16 | 37.83 | 41.70 |
| 16:57:16.199 | `0x03` | 0,1 | 19910 | 16:57:16 | 90.31 | 99.55 |
| 16:57:16.468 | `0x03` | 0,1 | 19570 | 16:57:16 | 88.77 | 97.85 |
| 16:57:18.015 | `0x03` | 0,1 | 20110 | 16:57:18 | 91.22 | 100.55 |
| 16:57:19.355 | `0x23` | 0,1,5 | 20110 | 16:57:19 | **91.22** | **100.55** |
| 16:57:20.267 | `0x23` | 0,1,5 | 20110 | 16:57:19 | 91.22 | 100.55 |

The first frame is stable **and** removed (bit 7): a stored record still being re-advertised 2 min 38 s after it was taken, which is the observation behind the staleness follow-up. Frames 2 to 5 are the live measurement settling as I step on. The last two are the same stable record broadcast twice 0.9 s apart, byte-identical.

Bit 0 is set on every frame of this session and clear on every frame of the kg session. Bit 2 never appears anywhere.

### Three things the log settles

1. **The unit flag.** `0x23` at 201.10 lb, matching the scale's display.
2. **Finding 4 — the Mi vendor history characteristic is present.** `00002a2f0000351221180009af100700` at handle `0x22`, alongside `1531` / `1532` / `1542` / `1543`. So a post-discovery re-resolution finds it and does not fall through to Standard GATT on this device.
3. **The local name is intermittent.** `name="MI SCALE2"` on most advertisements, `(none)` at `16:57:20.267` where the scan response did not arrive. That is exactly the moment Standard GATT takes the device and the GATT read fails on the missing `2a9c` / `2a9f`. It is also why the adapter matches on the exact name *and* the Huami fingerprint rather than either alone.

<details>
<summary>Full log (ESPHome proxy, DEBUG, ble-scale-sync 1.24.0)</summary>

```
2026-08-24 16:57:10.792 [BLE:debug] BLE handler: esphome-proxy
2026-08-24 16:57:11.064 [BLE] ESPHome proxy X.X.X.X:6053 connected
2026-08-24 16:57:11.066 [BLE] ESPHome proxy X.X.X.X:6053 initialized, BLE advertisement subscription active
2026-08-24 16:57:11.066 [BLE] ESPHome proxy connected at X.X.X.X:6053
2026-08-24 16:57:11.067 [BLE] ESPHome proxy transport ready (broadcast + GATT). Broadcast adapters: Eufy Smart Scale P2/P2 Pro, QN Scale, Xiaomi Mi Scale 2, Xiaomi Mijia Scale S800. GATT adapters (connected on demand): Senssun Fat Scale, Renpho ES-WBE28, Renpho R-MSC04, Renpho ES-26BB, Beurer BF720/BF105, Yunmai, Beurer / Sanitas, Sanitas SBF72/73, Soehnle Shape/Style, Medisana BS44x, Trisa, ES-CS20M, Exingtech Y1, Excelvan CF369, Hesley, Koogeek-S1, Inlife, Digoo, 1byone (Eufy), 1byone Scale (new), Active Era BS-06, Robi S9, Hutbit, MGB (Swan/Icomon/YG), Hoffen BS-8107, Standard GATT (BCS/WSS).
2026-08-24 16:57:14.421 [BLE:debug] Advert: [70:87:9E:ED:E5:E7] name="MI SCALE2" uuids=[181d] manufacturerData={0x0157: 70879eede5e7} serviceData={181d: a38e4eea07081810372b}
2026-08-24 16:57:14.422 [BLE:debug] Xiaomi Mi Scale 2 matched at 70:87:9E:ED:E5:E7 but broadcast frame is not stable yet
2026-08-24 16:57:14.423 [BLE:debug] Xiaomi Mi Scale 2 matched at 70:87:9E:ED:E5:E7 but broadcast frame is not stable yet
2026-08-24 16:57:14.856 [BLE:debug] Xiaomi Mi Scale 2 matched at 70:87:9E:ED:E5:E7 but broadcast frame is not stable yet
2026-08-24 16:57:15.764 [BLE:debug] Advert: [70:87:9E:ED:E5:E7] name="MI SCALE2" uuids=[181d] manufacturerData={0x0157: 70879eede5e7} serviceData={181d: 039420ea070818103910}
2026-08-24 16:57:15.764 [BLE:debug] Xiaomi Mi Scale 2 matched at 70:87:9E:ED:E5:E7 but broadcast frame is not stable yet
2026-08-24 16:57:16.199 [BLE:debug] Advert: [70:87:9E:ED:E5:E7] name="MI SCALE2" uuids=[181d] manufacturerData={0x0157: 70879eede5e7} serviceData={181d: 03c64dea070818103910}
2026-08-24 16:57:16.199 [BLE:debug] Xiaomi Mi Scale 2 matched at 70:87:9E:ED:E5:E7 but broadcast frame is not stable yet
2026-08-24 16:57:16.468 [BLE:debug] Advert: [70:87:9E:ED:E5:E7] name="MI SCALE2" uuids=[181d] manufacturerData={0x0157: 70879eede5e7} serviceData={181d: 03724cea070818103910}
2026-08-24 16:57:16.468 [BLE:debug] Xiaomi Mi Scale 2 matched at 70:87:9E:ED:E5:E7 but broadcast frame is not stable yet
2026-08-24 16:57:18.015 [BLE:debug] Advert: [70:87:9E:ED:E5:E7] name="MI SCALE2" uuids=[181d] manufacturerData={0x0157: 70879eede5e7} serviceData={181d: 038e4eea070818103912}
2026-08-24 16:57:18.015 [BLE:debug] Xiaomi Mi Scale 2 matched at 70:87:9E:ED:E5:E7 but broadcast frame is not stable yet
2026-08-24 16:57:19.355 [BLE:debug] Advert: [70:87:9E:ED:E5:E7] name="MI SCALE2" uuids=[181d] manufacturerData={0x0157: 70879eede5e7} serviceData={181d: 238e4eea070818103913}
2026-08-24 16:57:19.355 [BLE:debug] Xiaomi Mi Scale 2 matched at 70:87:9E:ED:E5:E7 but broadcast frame is not stable yet
2026-08-24 16:57:19.581 [BLE:debug] Xiaomi Mi Scale 2 matched at 70:87:9E:ED:E5:E7 but broadcast frame is not stable yet
2026-08-24 16:57:20.050 [BLE:debug] Xiaomi Mi Scale 2 matched at 70:87:9E:ED:E5:E7 but broadcast frame is not stable yet
2026-08-24 16:57:20.267 [BLE:debug] Advert: [70:87:9E:ED:E5:E7] name=(none) uuids=[181d] manufacturerData={0x0157: 70879eede5e7} serviceData={181d: 238e4eea070818103913}
2026-08-24 16:57:20.268 [BLE] Matched: Standard GATT (BCS/WSS) (70:87:9E:ED:E5:E7); opening GATT via ESPHome proxy
2026-08-24 16:57:21.833 [BLE:debug] ESPHome connected to 70:87:9E:ED:E5:E7 (mtu=27)
2026-08-24 16:57:21.833 [BLE:debug] ESPHome GATT char 00002a0000001000800000805f9b34fb handle=0x3 props=0xa cccd=none
2026-08-24 16:57:21.833 [BLE:debug] ESPHome GATT char 00002a0100001000800000805f9b34fb handle=0x5 props=0xa cccd=none
2026-08-24 16:57:21.833 [BLE:debug] ESPHome GATT char 00002a0200001000800000805f9b34fb handle=0x7 props=0x2 cccd=none
2026-08-24 16:57:21.834 [BLE:debug] ESPHome GATT char 00002a0400001000800000805f9b34fb handle=0x9 props=0x2 cccd=none
2026-08-24 16:57:21.834 [BLE:debug] ESPHome GATT char 00002a0500001000800000805f9b34fb handle=0xc props=0x22 cccd=0xd
2026-08-24 16:57:21.834 [BLE:debug] ESPHome GATT char 00002a2500001000800000805f9b34fb handle=0x10 props=0x2 cccd=none
2026-08-24 16:57:21.834 [BLE:debug] ESPHome GATT char 00002a2700001000800000805f9b34fb handle=0x12 props=0x2 cccd=none
2026-08-24 16:57:21.834 [BLE:debug] ESPHome GATT char 00002a2800001000800000805f9b34fb handle=0x14 props=0x2 cccd=none
2026-08-24 16:57:21.834 [BLE:debug] ESPHome GATT char 00002a2300001000800000805f9b34fb handle=0x16 props=0x2 cccd=none
2026-08-24 16:57:21.834 [BLE:debug] ESPHome GATT char 00002a5000001000800000805f9b34fb handle=0x18 props=0x2 cccd=none
2026-08-24 16:57:21.835 [BLE:debug] ESPHome GATT char 00002a2b00001000800000805f9b34fb handle=0x1b props=0xa cccd=none
2026-08-24 16:57:21.835 [BLE:debug] ESPHome GATT char 00002a9e00001000800000805f9b34fb handle=0x1d props=0x2 cccd=none
2026-08-24 16:57:21.835 [BLE:debug] ESPHome GATT char 00002a9d00001000800000805f9b34fb handle=0x1f props=0x20 cccd=0x20
2026-08-24 16:57:21.835 [BLE:debug] ESPHome GATT char 00002a2f0000351221180009af100700 handle=0x22 props=0x18 cccd=0x23
2026-08-24 16:57:21.835 [BLE:debug] ESPHome GATT char 000015310000351221180009af100700 handle=0x26 props=0x18 cccd=0x27
2026-08-24 16:57:21.835 [BLE:debug] ESPHome GATT char 000015320000351221180009af100700 handle=0x29 props=0x4 cccd=none
2026-08-24 16:57:21.835 [BLE:debug] ESPHome GATT char 00002a0400001000800000805f9b34fb handle=0x2b props=0x1a cccd=0x2c
2026-08-24 16:57:21.836 [BLE:debug] ESPHome GATT char 000015420000351221180009af100700 handle=0x2e props=0x1a cccd=0x2f
2026-08-24 16:57:21.836 [BLE:debug] ESPHome GATT char 000015430000351221180009af100700 handle=0x31 props=0x1a cccd=0x32
2026-08-24 16:57:21.837 [BLE:debug] Looking for notify=00002a9c00001000800000805f9b34fb, write=00002a9f00001000800000805f9b34fb
2026-08-24 16:57:21.839 [BLE] ESPHome proxy X.X.X.X:6053 disconnected
2026-08-24 16:57:21.839 [Sync] Required characteristics not found for adapter "Standard GATT (BCS/WSS)". Notify (00002a9c00001000800000805f9b34fb): false, Write (00002a9f00001000800000805f9b34fb): false. Discovered: [00002a0000001000800000805f9b34fb, 00002a0100001000800000805f9b34fb, 00002a0200001000800000805f9b34fb, 00002a0400001000800000805f9b34fb, 00002a0500001000800000805f9b34fb, 00002a2500001000800000805f9b34fb, 00002a2700001000800000805f9b34fb, 00002a2800001000800000805f9b34fb, 00002a2300001000800000805f9b34fb, 00002a5000001000800000805f9b34fb, 00002a2b00001000800000805f9b34fb, 00002a9e00001000800000805f9b34fb, 00002a9d00001000800000805f9b34fb, 00002a2f0000351221180009af100700, 000015310000351221180009af100700, 000015320000351221180009af100700, 000015420000351221180009af100700, 000015430000351221180009af100700]
2026-08-24 16:57:26.839 [BLE:debug] ESPHome GATT disconnect for 70:87:9E:ED:E5:E7 ignored: sendMessage timeout waiting for BluetoothDeviceConnectionResponse
2026-08-24 16:57:21.837 [BLE:debug] Looking for notify=00002a9c00001000800000805f9b34fb, write=00002a9f00001000800000805f9b34fb
2026-08-24 16:57:21.839 [BLE] ESPHome proxy X.X.X.X:6053 disconnected
2026-08-24 16:57:21.839 [Sync] Required characteristics not found for adapter "Standard GATT (BCS/WSS)". Notify (00002a9c00001000800000805f9b34fb): false, Write (00002a9f00001000800000805f9b34fb): false. Discovered: [00002a0000001000800000805f9b34fb, 00002a0100001000800000805f9b34fb, 00002a0200001000800000805f9b34fb, 00002a0400001000800000805f9b34fb, 00002a0500001000800000805f9b34fb, 00002a2500001000800000805f9b34fb, 00002a2700001000800000805f9b34fb, 00002a2800001000800000805f9b34fb, 00002a2300001000800000805f9b34fb, 00002a5000001000800000805f9b34fb, 00002a2b00001000800000805f9b34fb, 00002a9e00001000800000805f9b34fb, 00002a9d00001000800000805f9b34fb, 00002a2f0000351221180009af100700, 000015310000351221180009af100700, 000015320000351221180009af100700, 000015420000351221180009af100700, 000015430000351221180009af100700]
2026-08-24 16:57:26.839 [BLE:debug] ESPHome GATT disconnect for 70:87:9E:ED:E5:E7 ignored: sendMessage timeout waiting for BluetoothDeviceConnectionResponse
```

</details>

KristianP26 (approved):
The split into a dedicated adapter is the right shape, and the revised flag mapping is correct. I verified it rather than reading it.

Mutation check on the merged tree:

- `isLbs` back to `flags & 0x04` (the original mapping): 2 failed, with `expected { weight: 100.55 } to deeply equal { weight: 91.217425607 }`. That 100.55 is precisely the wrong number from the first revision, so the captured pounds frame now pins it.
- stable gate removed: 1 failed.

Your capture settles it independently of the SIG argument: bit 0 is clear on every frame of the kg session and set on every frame of the pounds session, and bit 2 never appears at all. Splitting the adapter also resolves my other two findings by construction, since MiScale2Adapter keeps its weight-only v2 behaviour untouched and `isComplete` can now be true, so a stable frame publishes immediately instead of waiting out the composition grace window.

Full suite on `dev` plus this branch: 132 files, 2238 tests, tsc, eslint and prettier clean. Registry is 31 entries, which lines up with the doc counts in the PR.

One note for later, not a blocker: the timestamp in bytes 3 to 9 is deliberately dropped, and the comment explains why. That is the right call for now, but it is also what would let a re-advertised stored record (your `0xa3` frame, stable and removed together, 2 min 38 s old) be recognised as historical rather than just filtered on the removed bit.

Issue #335, opened 2026-08-21
Beurer BF800: unlock write [f7 01] is never answered — reproduced on both the BlueZ and the ESPHome-proxy transport, with pairing agent, provisioning and pairing mode all ruled out


martingebert9428:
## Summary

A Beurer BF800-B connects reliably, exposes its full GATT tree, accepts the CCCD
write and the unlock write without error — and never sends a single
notification. The handler repeats `[f7 01]` on `ffe1` every 5 s with no state
progression until the scale drops the link.

I have systematically eliminated the transport, the BlueZ pairing agent, user
registration on the scale, and the scale's pairing mode. Behaviour is
byte-identical in every configuration. I am filing this as a possible
device-support gap for the BF800 specifically, and I am happy to run patched
builds — the device and a working debug setup are available.

## Environment

- BLE Scale Sync **1.24.0** (image v1.24.0 @ 3cce009), Home Assistant add-on
  (also reproduced on 1.23.0)
- `custom_config: true`, config at `/share/ble-scale-sync/config.yaml`
- Host: Home Assistant OS on a **Raspberry Pi 5** (built-in adapter, `hci0`)
- Transport A: `node-ble (BlueZ D-Bus)` — the local Pi 5 adapter
- Transport B: `esphome-proxy` — generic ESP32, ESPHome 2026.7.4, esp-idf
- Distance to the scale during all BlueZ runs: about 1 m
- Home Assistant's own Bluetooth integration was not holding the adapter

## Device

| Scale | MAC | Matched adapter |
|---|---|---|
| Beurer BF800-B | `00:13:04:10:1D:FC` | `Beurer / Sanitas` |

Advertisement:

```
addrType=public flags=06
uuids=[0000ffe0-..., 0000fe1b-..., 0000180a-...]
```

## GATT tree as discovered

```
1800  2a00  read
      2a01  read
1802  b1e0  read, write
1805  b1e3  read, notify
      b1e4  read, notify
ffe0  ffe1  read, write-without-response, write, notify
      ffe2  read, write-without-response, write, notify
180a  2a23..2a29  read   (Device Information)
```

## What works

- Discovery and MAC match
- GATT connect (0.8 s to 9.6 s, varies)
- Full service discovery
- Adapter selection: `Matched adapter: Beurer / Sanitas`
- CCCD subscribe on `ffe1` — no error
- Unlock write on `ffe1` — no error on the BlueZ transport

## What fails

```
[...]

PR #354, opened 2026-08-25
fix(beurer): decode impedance from body composition measurement frames

JamieSBenson:
## Summary

The BF720/BF105 adapter always reported impedance as 0 Ω because `parseBodyComposition()` skipped over the impedance field (off += 2) without reading it, and `buildReading()` hardcoded `impedance: 0`. The offset computation was already dynamic and correct — walking the flags bitmask in spec order. The bug was simply that the value was never captured into cachedComp and never fed into the reading.

## Changes

Fix: read the 2-byte little-endian impedance (0.1 Ω resolution) when bit 9 (0x0200) is set, store it in CachedComp, and use it in buildReading() with a `?? 0` fallback when the flag is absent.

## Test Plan

- [x] Tests pass (`npm test`) <- except for existing UTC test, which appears to have been fixed in a more recent PR

## Notes

Verified against two real BF720 notifications (flags 0x0398):
 - Payload 1: impedance 0x13ec = 5100 → 510.0 Ω
 - Payload 2: impedance 0x1004 = 4100 → 410.0 Ω

Also verified the existing #168 BCS frame now decodes impedance 0x11a8 = 452.0 Ω (previously 0).


KristianP26 (changes requested):
Thank you for this, and welcome. The impedance decode is a genuine find and I want it: `parseBodyComposition()` really did walk past bit 9 with a bare `off += 2`, and `buildReading()` really did hardcode `impedance: 0`, so every BF720, BF788 and BF950 reading this project has ever produced carried a zero it did not have to. The offset walk was already correct, so the fix is exactly as small as you say it is, and your two payload decodes check out arithmetically.

What I cannot take is the rest of the PR. There are three unrelated changes here, and one of them breaks correctly configured users.

## Blocker 1: the `height_unit` change reverses a documented contract with no migration

`docs/guide/configuration.md` describes `height_unit` as "Used for height input in user profiles", and `config.yaml.example` says `cm | in (for height input below)`. So the documented meaning is that `height_unit: in` means **the number in your config is in inches**, and converting it at resolve time is the behaviour working as specified, not a regression.

This PR makes `height_unit` display-only and treats the stored value as always centimetres. For anyone who configured it as documented, that is catastrophic and silent:

```yaml
scale:
  height_unit: in
users:
  - height: 68        # 68 inches, as the docs instruct
```

Today that resolves to 172.7 cm. After this change it resolves to 68 cm, and their BMI goes from about 24 to about 160. Nothing warns them, nothing migrates them, and the number that changes is the one the whole body-composition calculation is built on.

The tell is in the diff itself: the test named `converts height from inches to cm when height_unit is in` is deleted and replaced with one asserting the opposite. That test was pinning the specification, so a change that has to remove it is a specification change, not a bug fix.

The case you describe, a user who writes 172 while setting `height_unit: in`, is a misconfiguration rather than a defect. I agree it produces a bad result, and I agree the current design invites it. But the fix for "the setting is easy to get wrong" cannot be to silently reinterpret every config that got it right.

If you want to pursue storing canonical centimetres, and there is a real argument for it, it needs to be its own PR with a migration: detect the ambiguous case at load time, warn, and either convert once and rewrite the config or refuse to start. Happy to review that on its own merits. I would rather it did not ride along with an unrelated fix.

## Blocker 2: `package.json` carries a local tooling artifact

```json
"allowScripts": {
  "esbuild@0.28.1": true,
  ...
}
```

That is an allow-scripts block from your local setup. It is not part of this project's toolchain and should not land in the manifest.

## Please split this

Three changes are bundled: the impedance decode, the height-unit reinterpretation, and the wizard conversion that only exists to support the second one. The first is ready. The second needs its own discussion. The third is meaningless without the second.

If you open a PR with just the `src/scales/beurer-bf720.ts` change and its tests, I will review it straight away and it will very likely go in as is.

## One thing to know about the impedance fix, which is not an objection

It changes exported numbers for existing users, and that is worth saying out loud in the commit message. `computeMetrics()` passes `reading.impedance` into `buildPayload()`, which uses it for any field the scale did not supply natively. With impedance stuck at 0 those fields came from the Deurenberg estimate; with a real impedance they come from the BIA equations. That is an improvement, but someone watching a trend will see a step change on the day they upgrade, and they deserve to find that in the changelog rather than wonder about it.

## Small notes for the split PR

- `resolveUserProfile(user, _scaleConfig)` leaves the whole `ScaleConfig` parameter dead. If the height change ever does land, drop the parameter rather than underscore it, and update the callers.
- Your branch is named `dev`, which makes rebasing awkward on both sides. A topic branch per change is easier for both of us.
- `dev` has moved a fair distance today, including a fix to the BF720 date-of-birth provisioning that touches this same file. Worth rebasing before you resubmit.

Nice catch on the actual bug. It has been sitting there since the adapter was written.

KristianP26:
Status: this still merges cleanly against `dev`, which has moved a long way today, so no rebase is forced on you. Worth knowing what changed under it though, because it touches the same file:

- `src/scales/beurer-bf720.ts` gained Register New User (User Control Point 0x01, `users[].beurer_register_new_user`) and a session-end warning for a consent the scale never answers. Both are in `onConnected` and the UCP response handler, not in `parseBodyComposition` or `buildReading`, which is why your change is unaffected.
- The date-of-birth provisioning is now pinned by a timezone test.

The review from earlier still stands and is what is blocking merge. To restate the short version, since the thread is long now:

- **The impedance decode is a real find and I want it.** It is correct, it is small, and the field genuinely was being skipped.
- **The `height_unit` change cannot go in as written**: it reinterprets a documented setting, so a config that followed the docs (`height: 68` with `height_unit: in`) silently becomes 68 cm. It needs its own PR with a migration.
- **`package.json` picked up an `allowScripts` block** from your local tooling, which should not land.

A PR with just the `beurer-bf720.ts` change and its tests will go in quickly. If it is easier, say so and I will cherry-pick that file myself and credit you, rather than making you redo the branch.

KristianP26:
@JamieSBenson I have taken the impedance decode to `dev` as its own commit, with you as co-author, rather than leave the bug in place while the rest of this PR is sorted out. Commit `b3239f8`.

Your find is bigger than it looks. The test suite already contained a capture with bit 9 set, `0x11a8`, 452.0 ohm at 79.96 kg, and nobody had ever looked at it because `buildReading` hardcoded the zero. So this has affected every BF720, BF788 and BF950 reading the project has produced. It also fixes half of a separate report on #335 where a BF915 was logging `0 Ohm` on an otherwise clean session.

I added the reasoning to the comment, because the original "unused, native comp" note was only half right: the fields the scale does **not** supply are derived downstream, and a real impedance moves those from the BMI estimate onto the BIA equations. Exported numbers will step-change once, which needs saying in the release notes.

The other two changes are unchanged in my view:

- **`height_unit`** reinterprets a documented setting. A config that followed the docs with `height: 68` and `height_unit: in` currently means 172.7 cm and would silently become 68 cm. If you still want it, it needs its own PR with a migration and the existing contract test kept or deliberately replaced, and I am happy to review that.
- **`package.json`** picked up a lavamoat `allowScripts` block from your local tooling.

Closing this one since the part we agreed on has landed. Open a fresh PR for the height change whenever suits, and thanks, this was a good catch.

PR #355, opened 2026-08-25
Fix/salter clock

albert-canfield:
Summary
Follow-up to #341, from further hardware work on the same SA00656 unit. One new finding, plus the two minor findings from that review which did not make the merge.

The new one is the important part: this scale stores nothing at all until its clock has been set. An adapter that never writes the clock therefore reads empty slots forever on hardware that is working perfectly. A battery change is enough to cause it, and it presents as "the scale reports nothing" with no error to diagnose it by.

Measured on the unit after a battery pull, with its clock counting up from zero: fourteen slot sweeps either side of a real 88.5 kg weigh-in came back empty, and every 08 <i> counter read 0, on a scale answering every other command normally. Writing 01 <unix u32 LE> fixed it outright. The next weigh-in committed immediately and read back at age zero with its composition intact:

scale clock 1787593317
counter idx 1 -> 1 ; counter idx 7 -> 1
slot 1:  87.0 kg  fat 30.2%  water 48.9%  bmi 30.1        0s old
slot 7:  89.9 kg  (all zeros)                         13804s old
That run also answers the "before or after connectable" question from the #341 review. The record for the weigh-in in progress is written promptly and is readable at age zero on the next connect, so the current weigh-in is not raced. The risk of reporting a previous weigh-in comes from a backlog of uncollected records, which is what the age gate handles. MAX_RECORD_AGE_SEC is left at the merged five minutes.

Changes
Set the clock (01 <unix u32 LE>) when the scale is found with an unset one. Once per session, driven off the clock reply already in buildAck, and re-armed by onSessionEnd so a write that does not take is retried on the next connection. A clock that is already set is left alone: the vendor app writes local time into that UTC field and the scale drifts about 24 seconds an hour, so the two clocks disagree by an hour or more, and nothing here needs them to agree because every record is judged against the scale's own. Re-syncing would move the time base the vendor app's history is dated against. qn-scale.ts (0x20) and mgb.ts both re-sync every session; this is narrower on purpose. It is the adapter's only write that changes anything on the device, and the clear (0a) is still never emitted.
Reject a decoded weight above 300 kg, matching koogeek-s1.ts. The field is a bare u16, so unbounded it accepts up to 6553.5 kg, and any 20-byte frame whose first byte is under 8 is treated as a record. No floor is added, since the 7.4 kg fixture shows light loads are legitimate. Review finding 5 on #341.
Drop the bare charUuids: [CHR_FFC1] claim, which hits with no service or name qualifier at priority 125. Of the two options offered in that review this takes the first, since the name and both service forms already cover every observed advertisement, and a unit advertising neither never reaches discovery as a Salter anyway. Review finding 6 on #341.
Correct three prose blocks in the class comment that describe a record store model the hardware has since disproved. They call the store a rotating buffer that "rotates by itself" and state that the newest record is necessarily the current weigh-in. Reads are in fact stable and non-destructive (index 1 returned byte-identical data on fourteen consecutive reads), nothing the adapter does drains the store, and index 1 behaved as a queue head holding the oldest uncollected record: a weigh-in taken with nothing connected pushed 08 01 from 4 to 8 while index 1 kept returning a record 334 seconds old. One session exported a record stamped 6720 seconds before the scale's own clock, which is exactly what MAX_RECORD_AGE_SEC exists to stop. Left uncorrected, the next reader could reasonably conclude that gate is redundant.
Note the clock behaviour in the supported-scales row.
Test Plan
 Tests pass (npm test)
 Lint clean (npm run lint)
 TypeScript compiles (npx tsc --noEmit)
 Tested manually (describe below)
Full suite green on top of dev: 134 files, 2328 tests. Seven tests added, taking the Salter file from 36 to 43.

Four cover the clock write: that it writes when the clock is unset and that the timestamp it sends is the current time, that a set clock is left alone, that the 01 00 ack chains back into a clock read so the walk still starts inside the same cycle, and that it fires once per session but re-arms after onSessionEnd. Two cover the weight ceiling, at 6553.5 kg and just over the bound at 300.1 kg, alongside a check that the 7.4 kg fixture still parses. The characteristic-claim test is inverted to assert the bare FFC1 hit is refused, with a companion asserting the post-discovery service path still resolves with no name present.

The existing test that enumerates every write the adapter can produce and asserts none begins with 0a now covers the new paths too.

Hardware testing was on a SALTER-SA00656-BK, before and after a battery pull, which is what produced the unset-clock case in the first place.

Notes
Findings 5 and 6 above were both raised in the #341 review and are still present on dev. I have assumed they were lost when the staleness work was folded into the merge rather than deliberately dropped. If either was a deliberate call, say so and I will drop that hunk.

Same offer on the clock write: it is deliberately narrow, only touching a clock that has never been set. If you would rather it matched QN and MGB and re-synced every session, that is a small change.

Still inferred rather than measured: normalizesWeight. Every capture so far was taken with the scale displaying kg, so a record from a scale set to lb has never been seen. The inference rests on the record's companion fields being unambiguously metric, with bone mass in kg and BMR in kcal. Unchanged by this PR, and flagged because it is the first thing to check if a lb unit ever reports weights out by a factor of 2.2.

I also have a full DEBUG log of a real session, connect through to "Reading complete", which was asked for on #341 and never attached. Happy to add it here if useful.


KristianP26 (changes requested):
This is excellent work, and the finding at the centre of it is the kind that only comes from having the hardware and being willing to pull the batteries. "The scale stores nothing until its clock is set" explains a failure mode that would have read as a dead device forever, with no error to diagnose it by, and a battery change is enough to trigger it. Fourteen empty sweeps either side of a real weigh-in, with every `08 <i>` counter at zero, is a proper control.

It also corrects me. The prose I merged said the newest record IS this weigh-in. Your index-1-as-queue-head observation, and the session that exported a record stamped 6720 seconds before the scale's own clock, shows that is wrong. That is also the first independent confirmation that the age gate is load-bearing rather than defensive, which I had argued for on reasoning alone.

Two findings, one of which is a test rather than the code.

## 1. The weight-ceiling test passes for the wrong reason

`tests/scales/salter.test.ts:202`, `rejects an implausible weight rather than exporting it`, uses a bare `makeAdapter()` with no clock reply fed in. The no-clock gate rejects the record before `MAX_WEIGHT_KG` is ever consulted, so the test is green whatever the ceiling is set to.

Verified by mutation on the merged tree: setting `MAX_WEIGHT_KG = 100000` leaves the whole Salter file green, 43 of 43. The ceiling itself is fine, and I confirmed it works once the clock gate is satisfied:

```ts
const a = new SalterAdapter();
const over = Buffer.from(REC_876_BIA);
over.writeUInt16LE(3001, 6);                       // 300.1 kg
a.parseNotification(clockReply(over.readUInt32LE(2), 10));
expect(a.parseNotification(over)).toBeNull();      // passes; fails if the ceiling is raised
```

So the fix is to prime the clock in that test, the way the neighbouring tests do with `primed(...)`. Worth doing rather than waving through: an untested bound is one somebody removes during a cleanup with the suite still green, and this one is the only thing standing between a garbled 20-byte frame and a 6553.5 kg export.

I hit exactly this in my own work on this adapter yesterday, on a test of the no-clock gate that turned out to pass because the fixture was old enough for the age gate to reject it anyway. Same shape, and it is easy to miss because the assertion is genuinely true.

## 2. The clock is written in UTC, and your own note says the app writes local

`setClockCommand()` sends `Math.floor(Date.now() / 1000)`, which is UTC epoch seconds. Your comment two blocks up says the vendor app writes local time into that same field.

So after our write the scale holds UTC where the app would have held local, and every record taken afterwards is stamped on a different basis than the app's own history. Nothing in this project cares, exactly as you say, because every record is judged against the scale's own clock and the offset cancels. But the user's Salter app may well display those records shifted by their UTC offset, and that is a visible consequence of the one write we make to their device.

Since you have the unit and the app in front of you, this is a question rather than a change request: after the adapter sets the clock, does the app show subsequent weigh-ins at the right time? If it does, ignore this. If it shows them offset, matching the app's convention costs one line and keeps their history coherent.

I would rather ask than guess, because the argument cuts both ways: writing local time into a field the firmware may treat as UTC is its own kind of wrong, and you are the only one who can see which the device actually does.

## Everything else checks out

- The clock write is properly gated. Mutation-tested both ways: dropping the unset check fails 2 tests, removing the write entirely fails 3.
- Once per session with a re-arm on `onSessionEnd` is the right shape, and the reasoning for not retrying inside the session is sound.
- Leaving a set clock alone is the right call, and the drift argument for it is convincing.
- Dropping the bare `charUuids: [CHR_FFC1]` claim at priority 125 is the option I would have picked too.
- `MIN_PLAUSIBLE_CLOCK_SEC` at 2020-09-13 separates the two cases cleanly, and the "fifty years of uptime" note is the right way to justify a magic number.
- Full suite on `dev` plus this branch: 134 files, 2334 tests, tsc, eslint and prettier all clean.

One housekeeping note: this was opened against `main`, which made the diff 70 files because it swept in everything that landed on `dev` today. I have retargeted it to `dev` and it is now the three files you actually changed. Contributions land on `dev` and reach `main` only through a release promotion.

Fix the test and I will merge this. The clock question can be answered here rather than in code if the app turns out to be happy.

KristianP26:
Heads up: `dev` has moved a fair way since you opened this, and it now conflicts, in one place only.

`docs/guide/supported-scales.md` is the single conflicting file. Nothing to do with your clock work, it is the adapter-count prose and two new rows that landed today. Your `salter.ts` and its tests still apply cleanly.

While you are rebasing, the two review findings from earlier are the only things blocking merge:

1. The weight-ceiling test uses a bare adapter with no clock reply, so the no-clock gate rejects the record before `MAX_WEIGHT_KG` is consulted and the test is green whatever the ceiling is. Priming it the way the neighbouring tests do with `primed(...)` is the fix.
2. The clock is written as UTC epoch seconds while your own note says the vendor app writes local time into that field. That one is a question rather than a change request: after the adapter sets the clock, does the app show later weigh-ins at the right time?

No rush. The clock finding itself is the valuable part of this PR and it is not in question.

albert-canfield:
Follow-up from another round of hardware capture on the same SA00656, answering the open question from the #341 review and flagging one thing for a separate PR.

**Clock field is local time, now measured (not inferred).** A full PacketLogger trace of the vendor app on first contact shows it writing phone-local wall-clock as unix seconds — here `01 0E EE 8D 6A` = 1787686414, which is 19:33:34 read *as UTC* while the real wall clock was 19:33:34 BST, i.e. +3600 s off true UTC. It also re-writes the clock on every connect, even when it is already set. So a set clock is never a stable UTC value on a scale that is also used with the vendor app, which is exactly why this PR leaves a set clock alone and judges records against the scale's own clock. The "writes local time into that UTC field" line in the PR is now backed by the trace rather than inferred.

**Separate follow-up (not this PR): the record queue is position-indexed.** Byte three of `09 <slot> <n>` (and `0a <slot> <n>`) is a queue position, not a constant `00`. In the same trace the app drained a slot holding two weigh-ins in one session: `09 01 00` → 88.6 kg, `09 01 01` → 89.2 kg, then `0a 01 00` / `0a 01 01`. The adapter currently reads only `09 <index> 00` (position 0 = the oldest of each slot) and never clears, so behind a backlog it reads a stale record the age gate rejects while a fresher one sits unread at a higher position. A non-destructive `09 <slot> <count-1>` "read-newest" using the `08 <slot>` count fixes it. I'll open that as its own PR so this one stays scoped to the clock.

Nothing above changes this PR — it's all consistent with the merged design; posting it as the measured backing for the clock rationale and to record the follow-up.

KristianP26:
Measured answer on the clock field, and one that settles it: the app writing phone-local into a field read as unix seconds, plus re-writing it on every connect, means a set clock is never a stable UTC value anyway. Leaving a set clock alone and judging records against the scale's own clock is the right shape, and it is now backed rather than inferred. Thanks for going back to the analyser for it.

I pushed the one outstanding item myself rather than bounce the PR for three lines: the ceiling test now primes the clock on both adapters. Confirmed by mutation that it bites, `MAX_WEIGHT_KG = 100000` now fails that test where it used to leave the file green. Full suite, tsc, eslint and prettier all clean on the branch.

Merging. The position-indexed queue in #360 is a good find, and the read-newest walk is the right non-destructive answer, so take your time validating it on a real backlog.

PR #360, opened 2026-08-26
fix(salter): read the newest queued record per slot, not the oldest

albert-canfield:
## Summary
Stacks on #355. Another round of hardware capture on the same SA00656 decoded the record queue as **position-indexed**, which turns the driver's fixed `09 <index> 00` read into a latent staleness bug.

Byte three of `09 <slot> <n>` — and of `0a <slot> <n>` — is a **queue position**, not a constant `00`. In a vendor-app PacketLogger trace the app drained one slot holding two weigh-ins in a single session:

```
08 01      -> 08 01 02        slot 1 holds two records
09 01 00   -> 88.6 kg         position 0 = oldest
09 01 01   -> 89.2 kg         position 1 = newest
0a 01 00 / 0a 01 01           clears each after reading it
```

The adapter walks the slots reading `09 <index> 00` (position 0 = the **oldest** of each slot) and never clears. Once a user's slot holds more than one un-collected weigh-in, position 0 stays the oldest, `MAX_RECORD_AGE_SEC` rejects it as stale, and the reading that just happened — sitting at a higher position — is never seen. On a scale nothing ever drains, that is silent and permanent, and presents identically to the empty-slots failure #355 fixes for a different cause.

## Change
The walk now reads `08 <slot>` for each slot and fetches `09 <slot> <count-1>` — the newest position — when the count is non-zero, skipping empty slots straight to the next count. It is still non-destructive: **no `0a` is ever emitted**. Across slots the newest timestamp still wins and the age gate is unchanged. The class-comment prose and the walk tests are updated to match; a test asserts the newest position is fetched, never position 0.

The state machine per reply: clock → `08 0`; `08 <slot> <count>` → `09 <slot> <count-1>` if `count > 0`, else the next slot's `08`; record → the next slot's `08`; last slot ends the walk.

## Test Plan
- [x] Tests pass — Salter file 43 → 44 (added: fetches the newest position, never the oldest; the chaining test now covers count-then-fetch)
- [x] Full suite green on top of dev (post-#355 merge): 134 files, 2362 tests
- [x] Lint clean (`npx eslint`)
- [ ] **Driver path not yet re-checked on real hardware** — the position-indexing is validated from the vendor-app trace, and the walk logic from unit tests, but the assembled driver has not been run against the scale with a backlog since this change. **Draft for that reason.** I have the unit and can validate a two-record-backlog read before this is marked ready.

## Notes
- #355 is merged; this is now rebased onto dev as a single commit (2 files, +60/-22).
- Does not touch the clock write, the never-clear stance, or `MAX_RECORD_AGE_SEC`.

### DIFF
--- a/src/scales/salter.ts
+++ b/src/scales/salter.ts
@@ -26,19 +26,23 @@ const CHR_FFC1 = uuid16(0xffc1);
  * on FFC1.
  *
  * Not used here, but decoded and recorded so the next person does not have to:
- *   03 <i> <5 bytes>   write a user profile   04 <i>  read one back
- *   08 <i>             record-status probe    0a <i> 00  clear a record
+ *   03 <i> <5 bytes>   write a user profile      04 <i>  read one back
+ *   0a <i> <n>         clear the record at queue position n; answered `0a <i> 00`
  */
 const CMD_SET_CLOCK = 0x01; // `01 <seconds u32 LE>` -> `01 00`; only when unset
 const CMD_PING = 0x0b; // keepalive; the vendor app sends it before every poll
 const CMD_CLOCK = 0x02; // `02` → `02 <seconds u32 LE>`, the scale's own clock
-const CMD_FETCH = 0x09; // `09 <index> 00` → the 20-byte record in that slot
+const CMD_STATUS = 0x08; // `08 <index>` → `08 <index> <count>`, records queued in that slot
+const CMD_FETCH = 0x09; // `09 <index> <pos>` → the 20-byte record at that queue position
 
 /** `02 <u32 LE>`: opcode plus the clock, five bytes. */
 const CLOCK_REPLY_LEN = 5;
 
 /** `01 00`: the scale's acknowledgement of a clock write. */
 const SET_CLOCK_ACK_LEN = 2;
+
+/** `08 <index> <count>`: opcode, slot, queued-record count — three bytes. */
+const STATUS_REPLY_LEN = 3;
 
 /**
  * Below this the scale's clock is not a wall clock, so it has never been set.
@@ -110,15 +114,22 @@ const SWEEP_INTERVAL_MS = 3000;
 /**
  * How long the link is held open after the first record decodes.
  *
- * A sweep issues nine writes and their replies trickle back over a second or
- * two, so the window has to outlast a full cycle rather than ending the session
- * on the first frame that decodes. It also covers the case where the scale is
- * mid-reply when the first record lands.
+ * A sweep issues a count probe per slot and up to two fetches for each slot
+ * that holds records, and the replies trickle back over a second or two, so the
+ * window has to outlast a full cycle rather than ending the session on the first
+ * frame that decodes. It also covers the case where the scale is mid-reply when
+ * the first record lands.
  */
 const COMPLETION_HOLD_MS = 5000;
 
 /**
- * Measurement record: `<index> 00 | <unix seconds u32 LE> | <7 x u16 LE>`.
+ * Measurement record: `<slot> <position> | <unix seconds u32 LE> | <7 x u16 LE>`.
+ *
+ * The second byte echoes the queue position that was asked for: `09 01 01` was
+ * answered `01 01 ...` in the vendor-app trace. Every earlier capture read
+ * position 0, which is why the older fixtures all carry `00` there. It is
+ * logged, so a hardware run shows which end of a queue a record came from, but
+ * nothing is decided by it.
  *
  * The seven trailing fields are weight, body fat %, body water %, muscle mass %,
  * bone mass kg, BMR kcal and BMI — each scaled by ten except BMR, a whole number
@@ -129,6 +140,7 @@ const COMPLETION_HOLD_MS = 5000;
  * Only the weight is read; see the class comment for why the other six are not.
  */
 const RECORD_LEN = 20;
+const POSITION_OFFSET = 1;
 const TIMESTAMP_OFFSET = 2;
 const WEIGHT_OFFSET = 6;
 const WEIGHT_DIV = 10;
@@ -139,6 +151,11 @@ const TIMESTAMP_UNSET = 0xffffffff;
 /** True for a measurement record as opposed to a command echo. */
 function isRecord(data: Buffer): boolean {
   return data.length === RECORD_LEN && data[0] < RECORD_SLOTS;
+}
+
+/** `08 <slot + 1>`: the next slot's count, or nothing once the last is walked. */
+function statusOfNextSlot(slot: number): number[] | null {
+  return slot + 1 < RECORD_SLOTS ? [CMD_STATUS, slot + 1] : null;
 }
 
 /** `01 <unix seconds u32 LE>`: set the scale's clock to the host's time. */
@@ -160,9 +177,12 @@ function setClockCommand(): number[] {
  *
  *   -> 0b 00        <- 0b 00                  keepalive
  *   -> 02           <- 02 <clock u32 LE>      the scale's own clock
- *   -> 09 00 00     <- 00 00 ffffffff ...     slot 0, empty
- *   -> 09 01 00     <- ...                    each reply triggers the next
- *   -> 09 02 00     <- 02 00 <ts> <record>    slot 2, a real measurement
+ *   -> 08 00        <- 08 00 00               slot 0 count: empty, skip it
+ *   -> 08 01        <- 08 01 02               slot 1 count: two records queued
+ *   -> 09 01 01     <- 01 01 <ts> <record>    its newest position, count - 1
+ *   -> 09 01 00     <- 01 00 <ts> <record>    then its oldest, position 0
+ *   -> 08 02        <- 08 02 01               slot 2 count: one record, so
+ *   -> 09 02 00     <- 02 00 <ts> <record>    one fetch, its only position
  *
  * IT ALSO STORES NOTHING UNTIL ITS CLOCK IS SET, which is why the one write this
  * adapter makes that CHANGES anything on the device is `01 <unix u32 LE>`, and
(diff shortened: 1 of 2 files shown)


KristianP26:
#355 is merged, so this can be rebased down to the single commit whenever you are ready. Reviewed the second commit as it stands.

The position-indexed queue is a real find and the read-newest walk is the right answer to it: non-destructive, no `0a`, the age gate untouched, and it explains a failure that would otherwise present exactly like the empty-slots case you already fixed. The state machine reads cleanly and the walk tests cover the branches that matter, including the empty last slot ending it.

**One thing I want confirmed before this goes in, and you can answer it from the trace you already have.** The whole change rests on position 0 being the oldest. The evidence quoted is `09 01 00` giving 88.6 kg and `09 01 01` giving 89.2 kg, which shows the position byte is a real index but does not by itself say which end is newer. Records carry their own timestamp at bytes 2 to 5, so those two replies settle it outright. If `00` is stamped earlier than `01`, that is the confirmation and I would put it in the class comment where the current prose says so. If the direction is the other way, this change makes the bug worse rather than better, and it is silent either way, which is what makes it worth pinning rather than reasoning about.

Two smaller notes:

- `fetches the newest queued position, never the oldest` opens with a count of 1, where position 0 is both the newest and the oldest, so that assertion cannot fail for the reason the test name gives. The `08 01 03` case underneath is the one doing the work. Worth dropping the first or giving it a count above 1.
- The walk now costs two round trips per slot instead of one, so a full sweep goes from 8 replies to 16 on a scale that only answers when asked. Fine for a link that is held open, and skipping empty slots at the count stage claws some of it back, but if you see the sweep take noticeably longer on hardware that is why.

No objection to the design. Take the time on the two-record backlog test you mentioned, that is exactly the case that matters, and ping me when it is out of draft.

albert-canfield:
Marking this ready. Summary of the round that produced it, and where the validation stands, so you have the full picture.

**What the capture settled.** A full PacketLogger trace of the vendor app on the same SA00656 decoded the record store as a per-slot, position-indexed queue: byte three of `09 <slot> <n>` and `0a <slot> <n>` is a position, oldest at 0 and newest at `count - 1` (from `08 <slot>`). The app reads every position then clears each. That is the whole basis for this change — the adapter's `09 <index> 00` was reading the oldest of each slot, so behind any backlog it hands the age gate a stale record and never sees the weigh-in that just happened. The fix reads `08 <slot>` then `09 <slot> <count-1>`, still emitting no `0a`.

**Validation status, honestly.** The position-indexing and the walk are validated from the vendor trace and unit tests (full suite green, Salter file 43 → 44). The assembled driver has **not** been re-run against the scale with a two-record backlog since the change — I have the unit and can do that before you merge if you'd like; I left it out of scope only to keep the diff reviewable. Flagging it rather than implying a hardware pass it hasn't had.

**Two side findings from the same round, for context (no code here):**
- The clock field is *local* time, now measured, not inferred — details on #355. It re-writes on every vendor-app connect, so leaving a set clock alone is correct.
- The scale's body-composition (fat, water, muscle, bone) is a pure demographic calculation from weight + height + age + sex, with no impedance term: a one-weigh-in profile sweep reproduced every field from the profile alone to within rounding (fat-free mass to ±0.05 kg), and four separate weigh-ins fell on one line. So the adapter reporting weight only and leaving composition to the shared estimator is the right call — importing the scale's numbers would just be a second estimate, and a socks-on reading zeroes them. Not proposing any change here; recording it as backing for the current design.

Stacks on #355 (first commit is that PR). Happy to rebase to a single commit once #355 lands, or to run the backlog check on hardware first — your call on ordering.

albert-canfield:
Now that #355 is merged, I've rebased this onto dev — it's a single commit (2 files, +60/-22), no clock changes, mergeable. Only open item is the live two-record-backlog check on hardware, which I can run before merge if you want it verified first.

KristianP26:
Thanks for the rebase, and for flagging the hardware gap rather than glossing it. Single commit, CI green, and the walk itself reads well.

One thing is still open, and I want to be precise about why, because I think my last comment was easy to read as already answered.

The evidence for "position 0 is oldest" is still `09 01 00` -> 88.6 kg and `09 01 01` -> 89.2 kg. That proves the byte is a real index, which was never in doubt. It does not say which end is newer: it is equally consistent with the queue being newest-first and the person having weighed 89.2 then 88.6.

That matters more than usual here, because the two cases are opposites:

- position 0 is oldest -> today's code is broken behind a backlog and this fixes it
- position 0 is newest -> today's code is right and this reverses it

and both are silent. The records carry their own timestamp at bytes 2 to 5, so the two replies in your trace settle it in one line, without touching the scale.

**But you may not need to.** There is a version that does not depend on the answer at all: when `count > 1`, fetch position 0 **and** position `count - 1`, and let the newest-timestamp-wins comparison the adapter already does across slots pick between them. For `count == 1`, which is every slot on a scale that gets read daily, that is one fetch, exactly as now. Only a backlogged slot costs an extra round trip, and a backlogged slot is the entire reason this PR exists.

That fixes the reported bug, cannot regress if the ordering turns out to be the other way, and lets the class comment state what was measured rather than what was inferred. I would take that over the current version even with the timestamps confirmed, because it stops being an assumption.

Your call which way. Either the timestamp line from the trace, or the both-ends fetch, and I will merge. The hardware backlog check would be a nice confirmation on top but I would not hold the PR for it if the ordering is pinned.

Separately: the body-composition finding is a good one and worth its own issue rather than being buried here. A demographic calculation with no impedance term, reproduced to +/-0.05 kg on fat-free mass across four weigh-ins, is exactly the evidence that justifies the weight-only stance, and it will get lost in a merged PR thread.

albert-canfield:
Updated to the **both-ends fetch** you preferred, and answering the direction question from the trace — plus a full hardware pass on the assembled driver this time.

### The direction, settled by the records' own timestamps
From `salter 3 log.txt` (the same trace), the two records one slot held:

```
09 01 00 -> 01 00 d7 04 8e 6a 76 03 ...   stamp 1787692247   88.6 kg
09 01 01 -> 01 01 09 07 8e 6a 7c 03 ...   stamp 1787692809   89.2 kg
```

Position 1 is stamped 562 s after position 0, so on this unit **position 0 is the older end and `count - 1` the newest**. Record byte 1 echoes the position asked for (`01 01 …` for `09 01 01`); the adapter now logs it so a hardware run shows which end a record came from.

### The walk reads both ends, so it does not depend on that
I took your both-ends suggestion rather than trusting `count - 1`, for the reason you gave — if a unit ever queued the other way the failure would be the same one and silent. Per reply:

- clock → `08 00`
- `08 <slot> <count>` → `count 0`: next slot; else `09 <slot> <count-1>` (newest), and `count > 1` marks position 0 as owed
- record → the owed `09 <slot> 00` once, else the next slot's `08`
- the clock reply that restarts a walk, and `onSessionEnd`, drop the owed fetch, so a reply lost mid-slot cannot leave a fetch pending against the wrong slot

`count == 1` stays one fetch; only a backlogged slot costs the extra round trip. Still no `0a`, age gate and never-clear unchanged. Your smaller notes: the `count - 1` assertion is gone from the "newest" test (the backlog test uses counts 2 and 5 and asserts both fetches; a separate test asserts count 1 = one fetch); round trips measured from the trace are `08` ≈ 56 ms, `09` ≈ 150–180 ms.

### Hardware validation, the assembled driver this time
Ran the full app against a SALTER-SA00656-BK (`npx tsx src/index.ts`, native noble handler), not just the unit tests. All the cases that matter, each publishing exactly one reading — the newest — and age-gating the backlog behind it:

- **The bug, reproduced on 1.26.0:** with a backlog, position 0 returned the same 90.0 kg record every 3 s sweep, 46.7 h behind the scale's clock, age-gated each time, nothing published after a weigh-in.
- **Both-ends fix:** a real two-record backlog → newest published, older end age-gated (the old walk would have published nothing).
- **Unset clock after a battery pull:** clock written, the next weigh-in stored and published.
- **After a vendor-app sync moved the clock to local time (+3600 s):** fresh weigh-in still published, judged against the scale's own clock.
- **Two users:** routed by weight, per-user body composition, newest record only.

Squashed back to a single commit on `dev`. Body-composition-is-demographic will go in as its own issue as you suggested.

One separate finding, **not** part of this PR: on macOS/noble the scale powering off mid-session produces no `disconnect` event, so the session only ends on the idle timeout. A short `ble.session_timeout_sec` (~30 s) plus continuous mode recovers it; I'll write it up separately with the log excerpt.

KristianP26:
Pulled this down and went through the walk state machine line by line. It is good, and the way you settled the direction question is the right one: the two records' own timestamps, 562 s apart, are evidence from the trace rather than an assumption about the firmware, and reading both ends means the answer does not have to be right on every unit for the adapter to work. That is exactly what I was after.

Checks on my side: tsc, eslint and prettier clean, salter suite 47 passing, full suite 2484 passing.

Things I specifically looked at:

`oldestOwed` being a single flag rather than per-slot is fine, because every fetch is issued from the reply that decides it is owed, so only one slot is ever in flight. The restart path clears it and you have a test for that.

One narrow case is not covered and I do not think it needs to be, but I want it on the record so future me does not rediscover it as a bug. If a sweep restarts while a record reply from the abandoned walk is still in the air, that late reply can consume the flag that the new walk just set for a different slot. The result is that the new slot's position 0 is skipped for one cycle. It costs nothing: reads are non-destructive, the newest position was already fetched first so the reading that matters is in hand, and the next sweep is 3 s later. Not worth adding slot tracking for.

The class comment is doing real work now, particularly the paragraph explaining why only the two ends are read and not the middle. That is the part someone would otherwise "fix" later.

Merging. Thanks for going back to the trace instead of taking my suggestion at face value.

KristianP26 (approved):
Walk state machine checked line by line, full suite green. One benign restart race noted in the comment above, not worth guarding.

Issue #370, opened 2026-08-28
QN (FFE3-only firmware): the 0x1F reading ack never reaches the scale because session teardown nulls the context before the FFE3 fallback runs


alasano:
### Description

Same setup as #338 and #346, ESP32-S3 mqtt-proxy with autonomous connect and a QN scale on the older firmware that exposes FFE1/FFE3 and no FFF1/FFF2. Weigh-ins work and the exporters run, but every single reading ends with this line right after "Reading complete":

QN write failed on both 0000fff2... (Characteristic 0000fff2... not found) and 0000ffe3... (Cannot read properties of null (reading 'write')): [1f 05 15 10 49]

That [1f ...] frame is the ack the adapter sends after a stable reading so the scale knows it was received. It's the only write in the whole session that fails. The handshake writes earlier in the same session go out fine on FFE3 after FFF2 rejects, so the fallback works, it just never works for the ack.

From reading the code, the ack is sent fire-and-forget from parseWeightFrame, which then returns the reading. waitForReading picks that up and runs finishWith -> cleanup -> onSessionEnd in the same tick, and QN's onSessionEnd sets this.ctx to null. writeCmd checks this.ctx once at the top, awaits the FFF2 write, and when that rejects it reads this.ctx again for the FFE3 attempt. By then it's null, which is the TypeError in the log. On a scale that has FFF2 the ack goes out on the first attempt and none of this is visible, which is probably why it hasn't come up before.

The scale never gets told the reading landed, so it keeps advertising and the ESP32 keeps reconnecting to it, two or three times after every weigh-in on my unit, with a "Scale disconnected before reading completed" for each. That's the reconnect churn on top of every reading.

PR follows this issue.

### Steps to Reproduce

1. mqtt-proxy with auto_connect and a QN scale whose charmap has ffe3 but not fff2 (the "Looking for notify=... write=..." debug line shows which one you have)
2. Step on the scale and wait for "Reading complete"
3. The next debug line is the "QN write failed on both" for the [1f ...] frame, and the ESP32 reconnects to the scale a few seconds later

### Scale Brand / Model

Renpho ES-CS20M (QN, older firmware, FFE1/FFE3 only, proto 0x15)

### Platform

Docker (Linux, mqtt-proxy via ESP32-S3)

### Node.js Version

image default

### App Version

1.24.0, and the code path is unchanged on current dev

### Debug Logs

```
[...]

PR #371, opened 2026-08-28
fix(qn): hold the connection context across the FFE3 fallback so the 0x1F ack reaches the scale

alasano:
Full diagnosis in #370. The 0x1F ack after a stable reading is sent fire-and-forget from parseWeightFrame, and waitForReading ends the session in the same tick, which nulls this.ctx through onSessionEnd. writeCmd guarded this.ctx once and then read it again for the FFE3 fallback, so on scales that only expose FFE3 the FFF2 attempt rejected after the session had already ended and the fallback dereferenced null. The ack was lost on every reading, and the scale kept advertising and reconnecting because it never heard that the reading was received.

writeCmd now captures the context once and uses it for both attempts. The write is a fire-and-forget publish to the proxy, so sending it after onSessionEnd is fine, and I traced the ordering against the current watcher nesting: the FFF2 rejection's continuation runs before the reading's resolution propagates up through withIdleTimeout and withTimeout, so the FFE3 publish enters the MQTT client before the watcher's finally reaches mqttGattDisconnect, on both the host-initiated and autonomous paths. Capturing also pins the ack to its own session, so a replacement session's context can never be picked up by a stale fallback.

One test added: an FFE3-only context whose session ends right after parseNotification returns must still see the [1f 05 00 10 34] ack on FFE3. It fails with the writeCmd hunk reverted, no FFE3 ack is sent. Full suite, tsc, eslint and prettier clean on top of current dev.

### DIFF
--- a/src/scales/qn-scale.ts
+++ b/src/scales/qn-scale.ts
@@ -515,12 +515,17 @@ export class QnScaleAdapter
   }
 
   private async writeCmd(data: number[]): Promise<void> {
-    if (!this.ctx) return;
+    // Hold the context for both attempts. The 0x1F ack is issued from
+    // parseNotification, whose caller ends the session (onSessionEnd nulls
+    // this.ctx) in the same tick, so on FFE3-only scales the fallback would
+    // otherwise dereference null and the ack would never reach the scale.
+    const ctx = this.ctx;
+    if (!ctx) return;
     try {
-      await this.ctx.write(CHR_WRITE, data, false);
+      await ctx.write(CHR_WRITE, data, false);
     } catch (primaryErr: unknown) {
       try {
-        await this.ctx.write(CHR_WRITE_T1, data, false);
+        await ctx.write(CHR_WRITE_T1, data, false);
       } catch (altErr: unknown) {
         // Both write characteristics rejected. Logging this matters: a silent
         // return here is why a failed handshake looks identical to a scale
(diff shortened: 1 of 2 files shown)


KristianP26:
Checked this locally against dev and the diagnosis holds exactly as you wrote it.

What I wanted to be sure of was that the test really pins the bug rather than just passing alongside the fix, so I reverted `src/scales/qn-scale.ts` back to dev and kept your test. It fails on the assertion you would want it to:

```
AssertionError: expected undefined to be '0000ffe300001000800000805f9b34fb'
```

With the fix back in, the whole file is green (127 tests), and the full suite is 2484 passing, plus tsc, eslint and prettier clean.

Two things I checked beyond that:

The expected ack bytes are right. `[0x1f, 0x05, 0x00, 0x10, 0x34]` sums to 0x34 for `seenProtocolType` 0, and the frame in your log, `1f 05 15 10 49`, sums the same way for proto 0x15. So the test is consistent with the builder rather than hardcoding something that happens to pass.

The one design question was whether `writeCmd` should also copy the `this.ctx !== owner` guard that `writeAe01` has. It should not, and I want that written down somewhere so nobody adds it later "for consistency": in `writeAe01` that guard exists to stop a queued auth frame replaying onto a new link, but the 0x1F ack is by definition sent while the session is being torn down, so the same guard would put the bug straight back. The captured `ctx` already points at the specific old link, so there is no way for it to leak into a new connection.

The A2 echo path at the top of `parseWeightFrame` goes through the same `writeCmd`, so it gets the fix for free.

Merging. Thanks for tracking this one down to the tick ordering, that was the hard part.

KristianP26 (approved):
Verified locally: reverted the source fix, kept the test, confirmed it fails on the FFE3 assertion. Full suite green. Detailed notes in the comment above.

PR #373, opened 2026-08-29
feat(exporters): let the Garmin exporter upload the weight alone

O-Palda:
## Summary

Adds an opt-in `weight_only` setting for the Garmin exporter. When enabled, only weight is uploaded; Garmin receives the other body-composition values as absent rather than zero. Other exporters continue to receive the full measurement.

The setting is off by default, so existing configurations are unchanged.

```yaml
global_exporters:
  - type: garmin
    email: "${GARMIN_EMAIL}"
    password: "${GARMIN_PASSWORD}"
    weight_only: true
```

## Changes

- Supports `weight_only` in config, `GARMIN_WEIGHT_ONLY` for legacy `.env` users, and the Home Assistant add-on UI.
- Adds documentation and examples.
- Parses boolean environment values safely and reports only the values sent to Garmin.

Garmin may still calculate BMI from the uploaded weight and the height in the user profile; scale-derived metrics remain blank.

## Validation

- `npm test` (2506 tests)
- Python Garmin upload tests
- Lint, format, and TypeScript checks
- Manual verification of normal, weight-only, timestamp, `.env`, and add-on paths

### DIFF
--- a/.env.example
+++ b/.env.example
@@ -52,6 +52,14 @@ HEIGHT_UNIT=cm
 # EXPORTERS=garmin
 
 # ==========================================
+# GARMIN CONFIGURATION (only when garmin exporter is enabled)
+# ==========================================
+# Upload the weight alone. BMI, body fat, water, bone mass, muscle mass,
+# visceral fat, physique rating and metabolic age are left unset in Garmin
+# Connect. Other exporters are unaffected. Default: false
+# GARMIN_WEIGHT_ONLY=true
+
+# ==========================================
 # MQTT CONFIGURATION (only when mqtt exporter is enabled)
 # ==========================================
 # MQTT_BROKER_URL=mqtt://localhost:1883
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -70,3 +70,6 @@ jobs:
 
       - name: Run firmware unit tests
         run: python -m unittest discover -s firmware/tests -v
+
+      - name: Run Garmin uploader unit tests
+        run: python -m unittest discover -s garmin-scripts/tests -v
--- a/ble-scale-sync-addon/DOCS.md
+++ b/ble-scale-sync-addon/DOCS.md
@@ -64,6 +64,12 @@ To upload measurements to Garmin Connect:
 
 On first start the add-on authenticates with Garmin and stores the OAuth tokens under `/data/garmin-tokens` inside the container. Subsequent runs reuse those tokens, so your password is only used once.
 
+### Weight only
+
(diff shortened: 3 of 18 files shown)


KristianP26:
Hey, what's the reason for this?

KristianP26:
This is a well put together change. I went through all five layers it touches (env, config.yaml registry, add-on run.sh, the Python uploader and the docs) and they line up.

Verified locally on top of dev: tsc, eslint and prettier clean, Vitest 2506 passing (up 22), and the new Python suite 7 passing. I do not have `garminconnect` installed on this machine so `garmin_upload` would not import at first; CI installs it from requirements.txt, so that is my setup and not your change. I stubbed the module to confirm the tests actually exercise the branch rather than skipping.

The thing I like most here is not the feature, it is the bug you found on the way to it. `optionalBool` is right and the comment explaining why is right: `ExporterEntrySchema` is passthrough and `${ENV_VAR}` resolves to a string before the schema sees it, so `weight_only: "false"` under a bare `as boolean` cast reads as true, which is the exact inverse of what the user wrote, silently.

That bug is not limited to your field. Six others in `src/exporters/registry.ts` still take the unsafe cast:

- `mqtt.retain`
- `mqtt.ha_discovery`
- `ntfy.report_exports`
- `telegram.silent`
- `telegram.report_exports`
- `wger.sync_measurements`

`retain: "${MQTT_RETAIN}"` with `MQTT_RETAIN=false` currently retains. I am not asking you to widen the PR for that, the change is already at a sensible size and the helper you wrote is exactly what the others need. I will run it across the remaining six in a follow-up on dev and credit the diagnosis to this PR.

Two smaller things I checked and was happy with:

The `run.sh` guard. `opt_bool` output is interpolated unquoted into generated YAML, and the add-on schema does not bind a hand-edited options.json, so forcing anything that is not exactly `true` to `false` is the right call. That failure mode has bitten this repo before with YAML 1.1 parsing the generated file.

The uploader echoing `derived()` in its return value rather than the raw payload, so a weight-only run does not report metrics Garmin never received. Easy to miss and you did not.

Merging.

KristianP26 (approved):
All five layers checked, TS and Python suites green. The optionalBool find applies to six more fields; I will run it across them in a follow-up and credit this PR.

KristianP26:
Merged into dev as d270901, with you as the commit author.

It had to go in as a local squash rather than through the merge button: the branch touches `.github/workflows/ci.yml` (your new Garmin test job), and the API token this was merged with does not carry the `workflow` scope, so GitHub refused it with "refusing to allow an OAuth App to create or update workflow". Nothing to do with the change itself. Pushing over git worked because those credentials do have the scope.

Verified against current dev before pushing, since dev had moved on since your branch: tsc, eslint, prettier, 144 test files, 2536 tests, plus your 7 Python tests.

Closing this as merged rather than leaving it open. The optionalBool follow-up across the other six boolean fields is queued.

PR #374, opened 2026-08-30
feat(scales): add Xiaomi Body Composition Scale S400 adapter

gvarph:
## Summary

Adds support for the **Xiaomi Body Composition Scale S400** (MJTZC01YM, made by Yunmai; also the MJTZC03YM sibling). The S400 broadcasts encrypted MiBeacon v5 in FE95 service data like the S800, but under its own product ids (`0x30d9`, `0x3bd5`, `0x48cf`; `0x4b05` for MJTZC03YM) and a different object: `0x6e16` packs weight (0.1 kg), heart rate and impedance (0.1 Ω) into one uint32. The product ids and the byte layout are taken from Home Assistant's [`xiaomi-ble`](https://github.com/Bluetooth-Devices/xiaomi-ble) library (Apache-2.0; `devices.py` and the `obj6e16` parser in `parser.py`), which this adapter is a port of — credit to its authors for the reverse-engineering.

A barefoot weigh-in sends weight + 50 kHz impedance (+ heart rate), then a 250 kHz impedance frame; with socks only the weight frame. The 50 kHz value feeds the Xiaomi body-composition formulas shared with the Mi Scale 2 (`computeMiScaleComposition`, extracted from `mi-scale-2.ts`). Checked against the Mi Home report for one real weigh-in (86.1 kg / 512.5 Ω / 185 cm / 25 / male): the app says 22.7 % fat, 3.6 kg bone, 35.5 kg skeletal muscle; the Xiaomi formulas give 24.3 % / 3.3 kg / 33.9 kg, while the generic BIA coefficients gave 28.4 % / 2.6 kg. The app runs Yunmai's proprietary dual-frequency model, so exact agreement is not available; a test pins the distance on that weigh-in. Heart rate and the 250 kHz value have no field in `ScaleReading` and are logged only.

Verified live on my unit (product id `0x3bd5`, firmware model `yunmai.scales.ms104`): `85.7 kg, 50 kHz impedance 514 ohm, heart rate 74 bpm (profile 1)` → exported.

## Changes

- `src/scales/xiaomi-s400.ts` — new adapter (priority 205, `custom: true` because it shares `fe95` with the S800; matched by product id or the `Xiaomi Scale S400` name). Decodes `0x6e16`, classifies weight / impedance-high / step-off frames, warns once each for missing bind key, unknown MAC, and failed decryption; logs each distinct frame once.
- `src/scales/mi-scale-2.ts` — `computeMiScaleComposition()` extracted from `MiScale2Adapter.computeMetrics` (no behaviour change) so the S400 can reuse the Xiaomi formulas.
- `src/scales/mibeacon.ts` — MiBeacon helpers lifted out of `xiaomi-s800.ts` (`decryptMiBeaconV5`, `macFrameOrderFromFrame`, `normUuid`, `SVC_FE95`) plus an object iterator and a payload-offset helper that honours the capability byte. `xiaomi-s800.ts` imports and re-exports them, so its tests are untouched.
- `AdapterRuntimeConfig.scaleMac` (+ `run.ts`): the S400's measurement frames omit the MAC that the AES-CCM nonce needs and the scale is sleepy, so the adapter caches the MAC from the idle beacon and falls back to the configured `ble.scale_mac`.
- Registry + `registry-collision` / `adapter-resolution` fixtures (S400 vs S800 resolve correctly from service data alone and from the name alone).
- Docs: supported-scales row + Known Limitations entry, `bind_key` wording (S800 / S400), `config.yaml.example`, adapter count 32 → 33.

## Test Plan

- [x] Tests pass (`npm test`) — 2522 tests; new `tests/scales/xiaomi-s400.test.ts` (synthetic AES-CCM round-trips, MAC caching / `scale_mac` fallback, socks / high-impedance / reset frames, warnings) and `tests/scales/mibeacon.test.ts`. The S400 suite also replays the four public S400 fixtures from the [`xiaomi-ble`](https://github.com/Bluetooth-Devices/xiaomi-ble/blob/main/tests/test_parser.py) test-suite (Apache-2.0) so the decoder is provably byte-compatible with Home Assistant's.
- [x] Lint clean (`npm run lint`), `npm run format:check`
- [x] Tested manually — real S400 over a proxy transport, weight + impedance + heart rate decoded and exported (see above)

## Notes

- `ble.bind_key` stays a single global key (same field as the S800); one Xiaomi scale per instance, as today.
- The reference vectors in the test carry the (published) bind keys from the `xiaomi-ble` fixtures; happy to drop them and keep only the synthetic round-trips if you prefer the S800 test's "no real keys" convention strictly.


KristianP26:
This is the most careful adapter PR this repo has had in a while. Full suite is green here (145 files, 2523 tests), tsc, eslint and prettier clean.

The `mibeacon.ts` extraction is a straight move with a re-export left behind, so the S800 keeps working and the shared parts stop being copy-paste. Threading `scaleMac` through `AdapterRuntimeConfig` is additive and optional, which is the right shape for it.

The part I want to call out, because it is the opposite of a mistake I found elsewhere today: your `computeMetrics` actually runs a BIA model. Five existing adapters read an impedance off the wire and then quietly hand `{}` to `buildPayload`, which means their body fat is the BMI estimate no matter what the scale measured. I opened #386 for those. Yours does not have that problem, and the comment explaining why the Mi Scale 2 formula set was chosen over the generic coefficients, with the numbers you compared, is exactly the kind of note that stops someone "simplifying" it later.

Two things to fix, both small.

**The MAC preference order contradicts its own comment**

The class comment says the idle beacon's MAC is cached and `ble.scale_mac` is the fallback. The code does it the other way round:

```ts
const mac = frameMac ?? this.configuredMac ?? this.cachedMac;
```

A configured MAC therefore wins over one actually observed from the device. Normally they are the same value so nothing shows, but if someone pins a MAC with a typo, or the adapter sees the scale under an address that differs from what is configured, decryption fails even though the correct MAC is sitting in `cachedMac`. The warning they then get points at the bind key, which sends them looking in the wrong place.

`frameMac ?? this.cachedMac ?? this.configuredMac` matches the comment and prefers observed reality over configuration. Your test at line 162 passes either way because it does not set `scaleMac`, so a case with both set would be worth adding.

**One adapter count is still on the old number**

You updated four of the places that carry it, which is more than most PRs manage. `docs/public/logo.svg` is the one left: it still reads `32.11` and should be `33.11`. It is not a version, it is adapters dot exporters.

Heads up that #383 adds another adapter, so whichever of the two lands second will need to go to 34 across the board. I will sort that out at merge time rather than making you chase it.

One question. The fixtures come from the xiaomi-ble test suite rather than your own capture, which you say plainly and I am fine with, since those vectors came off real hardware and the licence direction works (Apache-2.0 into GPL-3.0 is fine, and you credited it in both files). But do you actually have an S400 in front of you? If you do, a single real weigh-in with `DEBUG=true` would confirm the end-to-end path, particularly that the idle beacon turns up before the measurement frames in practice, which is the assumption the whole nonce-caching design rests on. If you do not have one, say so and I will merge it as decoded-but-unverified-on-hardware and label it so the first user report is read in that light.

KristianP26:
I rebased this onto dev and pushed two follow-ups to your branch, so you do not have to chase either of them.

The second one is my fault. The Etekcity ESF-551 adapter landed on dev while this PR was open, which moved the adapter counts underneath you and put this branch into conflict on the docs. I have redone the count across all six places: registry 35, protocol adapters 34. The supported-scales table now carries both new rows.

The first is the MAC preference we discussed. It now reads:

```ts
const mac = frameMac ?? this.cachedMac ?? this.configuredMac;
```

which matches what your class comment already says. I added a test with both a cached MAC and a deliberately wrong configured one, and confirmed it fails on the old order, so it pins the behaviour rather than passing either way.

Everything green on the rebased branch: tsc, eslint, prettier, 146 test files, 2549 tests.

That clears both of my review points. The open question is still the hardware one: do you have an S400 in front of you? A single real weigh-in with DEBUG on would confirm the idle beacon arrives before the measurement frames, which is what the whole nonce-caching design rests on. If you do not have one, say so and I will merge it labelled as decoded but unverified on hardware, so the first user report gets read in that light.

gvarph:
Thanks for the rebase and for taking both fixes yourself; the order and the recount look right to me.

**Hardware** — yes, I have the scale in front of me: an S400 that identifies as MJTZC01YM with product id `0x3bd5` (its Mi Home record says model `yunmai.scales.ms104`, so that string exists in the wild alongside `ms103`). It has been my daily scale on this adapter for the past week: 19 weigh-ins between 2026-08-30 and today, every one decoded and exported (file, MQTT → Home Assistant, Garmin Connect), 16 barefoot with impedance and 3 in socks that correctly came through as weight-only. No missed or duplicated readings (with the Mi Home app closed; see below for what happens when it is not). Everything below was read over the `ha-bluetooth` transport from #375 (SLZB-MR5U → Home Assistant 2026.9.0 → websocket), but the frames are the raw FE95 service data HA relays, so the transport is incidental to what you asked.

Two `DEBUG` weigh-ins. The first was on the branch as reviewed, and I want to be straight about it: `scale_mac` was configured, so with the old ordering the configured MAC did the decrypting and the cache path was not exercised even though the beacon was there:

```
2026-08-30 09:11:13.205 Advert: [F8:83:06:4E:B6:7E] serviceData={fe95: 1059d53b0a7eb64e0683f8}            <- idle beacon, FC 0x5910: MAC included, cnt 0x0a
2026-08-30 09:12:22.819 Advert: [F8:83:06:4E:B6:7E] serviceData={fe95: 4859d53b0b14292b40f0cda8187877268f000000a16f90f7}   <- FC 0x5948: encrypted, no MAC, cnt 0x0b
2026-08-30 09:12:22.823 Xiaomi S400: 85.7 kg, 50 kHz impedance 514 ohm, heart rate 74 bpm (profile 1)
2026-08-30 09:12:23.553 Advert: [F8:83:06:4E:B6:7E] serviceData={fe95: 4859d53b0cea3cf917259bddd468d7d329000000307bc729}
2026-08-30 09:12:23.553 Xiaomi S400: 250 kHz impedance 464 ohm (profile 1)
2026-08-30 09:12:24.819 Advert: [F8:83:06:4E:B6:7E] serviceData={fe95: 1059d53b0c7eb64e0683f8}            <- idle beacon again, cnt 0x0c
```

The second is today with the corrected order (a local build carrying the same one-line change you pushed), same config, so this time the cached beacon MAC is what decrypted the frame:

```
2026-09-05 12:32:48.460 Advert: [F8:83:06:4E:B6:7E] serviceData={fe95: 1059d53b2a7eb64e0683f8}            <- idle beacon, MAC included, cnt 0x2a (~60 s before anyone stepped on)
2026-09-05 12:33:50.114 Advert: [F8:83:06:4E:B6:7E] serviceData={fe95: 4859d53b2b32311bfb89f73ca8401e41110000000bb0a3de}   <- cnt 0x2b, no MAC
2026-09-05 12:33:50.118 Xiaomi S400: 85.8 kg, 50 kHz impedance 509.8 ohm, heart rate 99 bpm (profile 1)
2026-09-05 12:33:50.118 Matched: Xiaomi Body Composition Scale S400 (F8:83:06:4E:B6:7E)
2026-09-05 12:33:50.120 Exporting to: file, mqtt, garmin...
2026-09-05 12:33:51.064 Advert: [F8:83:06:4E:B6:7E] serviceData={fe95: 4859d53b2c6406e3220706d5c3a2c6d03c000000a78131f0}
2026-09-05 12:33:51.065 Xiaomi S400: 250 kHz impedance 462.5 ohm (profile 1)
2026-09-05 12:33:52.856 Advert: [F8:83:06:4E:B6:7E] serviceData={fe95: 1059d53b2c7eb64e0683f8}
```

The counter tells the story on its own: with `scale_mac` still set to the true address either source would have decrypted, but the order is now cache first, so the beacon at `0x2a` is what fed the nonce for `0x2b`. The ordering assumption holds in practice, and with a bonus: the S400 emits the MAC-carrying idle beacon periodically while nobody is on it (both runs caught one about a minute before the weigh-in, and the frame counters are consecutive: beacon 0x2a, weight 0x2b, 250 kHz 0x2c), so the cache is warm before the first measurement frame rather than racing it. The frames above are ciphertext only; the bind key is never logged.

Two more observations from today that bear on the design, both from the frame counter (the byte after the product id, part of the CCM nonce):

- **The idle beacon is a passive re-broadcast, not a heartbeat with its own counter.** It always carries the counter of the last measurement frame (`0x2c` after the 12:33 weigh-in above) and repeats unchanged until the next one. So a stale cache cannot be told apart by counter, but it also never needs to be: the MAC in it is the only thing the adapter uses.
- **A weigh-in taken while the Mi Home app is open does not show up as MiBeacon frames.** I took one at 12:38 with the app in the foreground; the app tracked the weight live and stored the result, HA saw no FE95 frames at all, and the beacon was still `0x2c` when I probed HA's cache at 12:45 and again live at 12:49. Not a reception miss: three weigh-ins right after with the app closed consumed exactly `0x2d`–`0x32`, six frames, two per weigh-in (one in socks: `0x2f` weight-only, then `0x30` "stepped off", and both decoded). The likely mechanism is that the app holds a GATT connection and the scale streams the live measurement over that instead, i.e. a different protocol path rather than the same objects on another channel; I have not sniffed the connection to confirm, so I would not rule out an occasional broadcast in that state, only that none happened here. Either way it is the "keep the app closed while weighing" caveat in the docs, now with evidence, and it applies to every passive reader of this scale.

```
2026-09-05 12:51:37.835 Advert: [F8:83:06:4E:B6:7E] serviceData={fe95: 4859d53b2d4055cbf4d300113e948e620f000000b33c3303}   <- cnt 0x2d, 18 min after the last beacon; cache still warm
2026-09-05 12:51:37.836 Xiaomi S400: 85.9 kg, 50 kHz impedance 510.8 ohm, heart rate 84 bpm (profile 1)
2026-09-05 12:51:38.674 Xiaomi S400: 250 kHz impedance 463.4 ohm (profile 1)                              <- cnt 0x2e
2026-09-05 12:51:39.732 Advert: [F8:83:06:4E:B6:7E] serviceData={fe95: 1059d53b2e7eb64e0683f8}            <- beacon re-broadcasts cnt 0x2e
2026-09-05 12:52:07.426 Xiaomi S400: 86.2 kg, no impedance (profile 1)                                    <- cnt 0x2f, socks
2026-09-05 12:52:08.483 Xiaomi S400: person stepped off                                                   <- cnt 0x30
2026-09-05 12:52:37.659 Xiaomi S400: 85.8 kg, 50 kHz impedance 510.8 ohm, heart rate 72 bpm (profile 1)   <- cnt 0x31
2026-09-05 12:52:38.710 Xiaomi S400: 250 kHz impedance 463.4 ohm (profile 1)                              <- cnt 0x32
```

Body-composition data point for the formula note, same weigh-in in both (12:52:37 above, app closed while weighing and opened afterwards to sync): app 85.8 kg, 22.6 % fat (19.4 kg), 48.4 kg water (56.4 %), 3.6 kg bone, BMI 25.1 vs adapter 85.8 kg, 24.2 % fat (20.7 kg), 52.0 % water (44.6 kg), 3.3 kg bone, BMI 25.1. Muscle is not comparable: the app reports total muscle (62.8 kg) where the Mi Scale 2 formula reports skeletal muscle (34.0 kg). On Aug 30 the pair was app 22.7 % / 3.6 kg bone vs adapter 24.3 % / 3.3 kg, where the generic BIA coefficients would have given 28.4 % / 2.6 kg. That gap is what the comment in `computeMetrics` summarises.

KristianP26 (approved):
Approved after rebasing it myself and taking both review points. Full suite green locally on the rebased branch.

KristianP26:
This is the best hardware report this project has had. Thank you, and specifically for measuring the app-open case with the frame counter instead of asserting it.

Everything you found is recorded now, in the internal protocol notes rather than only in this thread, because three of the four points are load-bearing for the design and would otherwise be rediscovered the hard way.

**The idle beacon arrives before the measurement frame, not racing it.** That was the assumption the whole nonce-caching design rests on and the one I asked you to check. Consecutive counters, beacon 0x2a, weight 0x2b, 250 kHz 0x2c, about a minute apart, settle it.

**The beacon is a passive re-broadcast carrying the last measurement's counter.** I would have assumed it had its own. It means a stale cache cannot be spotted by counter, which is fine because the adapter only takes the MAC out of it, but anyone later tempted to use that counter for freshness now knows not to.

**Product id 0x3bd5 with model yunmai.scales.ms104.** Good to have a second model string confirmed in the wild rather than inferred from the id set.

**The Mi Home app takes the measurement.** Your counter continuity is the right way to prove that, and it is stronger than "we did not see anything": 0x2c still standing eleven minutes later, then 0x2d through 0x32 consumed exactly by the three later weigh-ins, two frames each. I have written the GATT-connection explanation down as your hypothesis rather than as fact, since you say plainly you did not sniff the connection. The docs already carried the "keep the app closed" caveat; now it has evidence behind it.

The body-composition comparison is the part that justifies the formula choice in writing: 22.6 against 24.2 percent versus the app, where the generic coefficients would have given 28.4. That is the difference between a defensible offset and a wrong number, and it now sits next to the code that makes the choice.

One thing you did that I want to name because it is easy to get wrong: the frames you posted are ciphertext only and the bind key never appears. That is exactly right for a public thread.

Separately, your follow-up commit on #375 listing which adapters the transport can serve did make it into that merge, so it is on dev. I checked the seven rows against the live registry rather than taking them on trust and they match exactly, including the S400 row you flagged as a judgement call. Keeping it was right.

Nothing outstanding from my side on either PR.

PR #375, opened 2026-08-30
feat(ble): add Home Assistant Bluetooth transport (ha-bluetooth)

gvarph:
## Summary

Adds `ble.handler: ha-bluetooth`: Home Assistant itself becomes the BLE radio. The app authenticates to HA's websocket API with a long-lived admin token and subscribes to `bluetooth/subscribe_advertisements`, so **every scanner HA already owns** — its local adapter, ESPHome Bluetooth proxies, SMLIGHT SLZB coordinators (native BLE proxy, HA ≥ 2026.8), Shelly devices — feeds the existing passive pipeline. Unlike pointing the app at an ESPHome node directly, nothing is taken away from HA (no "only one API subscription" problem), and it works for people whose only BLE receiver near the scale is something HA already uses.

Broadcast scales only: HA exposes no GATT path over this API, so GATT-only adapters are named at startup and warned about once per device instead of timing out silently.

Motivation: my server's Bluetooth is dead, and the only radio near the scale is a SMLIGHT SLZB-MR5U whose BLE proxy speaks a proprietary UDP protocol to HA. Going through HA covers that and every other HA proxy in one transport. Verified live with a Xiaomi S400 (weight + impedance + heart rate decoded and exported through this transport).

## Changes

- `src/ble/handler-ha-bluetooth/` — `client.ts` (websocket client on Node 22's global `WebSocket`, no new dependency: auth handshake, subscribe, reconnect with 2 s → 60 s backoff, ping/pong dead-link detection; a rejected token or non-admin user is a permanent error and never retried), `advert.ts` (HA advertisement → `BleDeviceInfo`; HA reports the address as the name of nameless devices, mapped to an empty name), `scan.ts` (`scanAndReadRaw` / `scanAndRead` / `scanDevices`, same shape as the esphome handler minus GATT), `watcher.ts` (`ReadingWatcher` implementing `Watcher` incl. `lastTransportActivityMs`, so `proxy_liveness_timeout_min` applies unchanged).
- HA replays its advertisement cache on subscribe; anything HA last saw more than 30 s ago is dropped so a restart cannot re-export the previous weigh-in.
- Config: `HaBluetoothSchema` (`url`, `token`, optional `source` scanner filter), handler enum + refine, `BleLoadedConfig` / `ResolvedRuntimeConfig` / `AppContext` / `ScanOptions` plumbing, `BLE_HANDLER=ha-bluetooth` override (accepted only when the block exists, like mqtt-proxy), restart-required diff with the token redacted, unknown-key reporting section, `transport-availability` maps.
- Wizard: fourth handler choice + prompts.
- Docs: new `docs/guide/ha-bluetooth.md` (sidebar entry), configuration table rows and proxy prose, `config.yaml.example` block, README bullet.

## Test Plan

- [x] Tests pass (`npm test`) — 2527 tests; new `tests/ble/ha-bluetooth/{client,advert,scan-watcher}.test.ts` (scripted fake WebSocket: handshake, auth_invalid / unauthorized / unknown_command, connect timeout, stale-cache filtering, source filter, ping-pong drop + reconnect, stop semantics; scan/watcher with a mocked client: broadcast/partial/grace/dedup/targetMac/GATT-warning paths) plus cases added to schema, unknown-keys, env-overrides, reload-diff, resolve-handler-key, transport-availability and wizard tests.
- [x] Lint clean (`npm run lint`), `npm run format:check`
- [x] Tested manually — HA 2026.8.3 with a SMLIGHT SLZB-MR5U proxy; a real weigh-in was matched, decoded and exported through this transport in continuous mode.

## Notes

- Requires Node ≥ 22 for the global `WebSocket` (already the `engines` floor).
- The HA add-on is unchanged: it can use this via `custom_config`. First-class add-on options could follow if wanted.
- `docker-entrypoint.sh` still runs the `btmgmt` adapter reset for this handler (harmless "reset failed" line, same as esphome-proxy today); left as-is to keep the diff focused.


KristianP26:
Went through this properly, because a fourth BLE transport is the kind of thing that is hard to walk back once people depend on it. It is in good shape.

Green locally on top of dev: tsc, eslint, prettier, 146 test files, 2527 tests. CI here only shows GitGuardian so far, the rest needs a maintainer to release the workflow run for a first-time contributor. I will do that.

Things I checked rather than assumed:

**The `time` field.** This was the one that worried me. If HA sent `time` as a monotonic clock, `this.now() - ad.time * 1000` would exceed `STALE_ADVERT_MS` on literally every advertisement and the transport would sit there silently receiving nothing, with your tests still passing because they build epoch stamps. So I read HA's `serialize_service_info` instead of guessing. It emits `service_info.time + time_diff` where `time_diff = round(time.time() - time.monotonic(), 2)`, computed once per subscription. `service_info.time` is monotonic and that offset converts it, so what reaches the wire is unix epoch and your comparison is correct.

**Secret handling.** `ble.ha_bluetooth.token` is in the `maskSensitive` list in `reload-diff.ts`, and the token is only ever sent, never logged. That is the thing most easily forgotten when a new credential is added and you did not forget it.

**The `WsLike` interface.** Declaring the slice of the WebSocket surface locally rather than pulling in the DOM lib, and injecting a factory for tests, is the right call and it is why `client.test.ts` can be as thorough as it is.

Two robustness points. Neither blocks the merge for me, but both are about the same failure shape, which is this transport going quiet without saying why.

**1. Clock skew reads as total silence.** The stale gate compares HA's wall clock against ours. If the two hosts are more than 30 s apart, and a Pi with no RTC that came up before NTP settled is exactly that case, every advertisement is dropped. The user sees three debug lines (`staleDropped <= 3`) and then nothing at all, with the subscription reporting itself as healthy. Since you already count `staleDropped`, it is cheap to say so: if some window passes with a nonzero drop count and zero advertisements delivered, warn once and print the measured offset. That turns a mystery into a one-line diagnosis.

**2. A permanent error after the first connection leaves the transport dead but quiet.** `auth_invalid` and a refused subscription both set `stopped = true`. On the initial connect that surfaces properly, because `start()` rejects. On a reconnect it lands in the `scheduleReconnect` catch and comes out as a single `bleLog.warn('Home Assistant reconnect failed: ...')`, after which nothing ever reconnects. Someone who rotates their long-lived token six months from now gets one warning line in a running log and a scale that silently stops working. Worth an `error` level, and worth saying plainly that the transport has given up rather than just that one attempt failed.

Both are small enough that I am happy to take them here or in a follow-up. Your call. Everything else I looked at, the backoff reset on a successful subscribe, `unref` on both timers, the superseded-socket guard in the close handler, per-subscriber try/catch in `dispatch`, and the specific messages in `describeSubscribeError` for the admin-token and old-HA cases, is exactly what I would want.

One question about scope, for the docs rather than the code. This is passive-only, so scales needing a GATT connection cannot use it. `docs/guide/ha-bluetooth.md` says so, but the supported-scales table does not mark which adapters are broadcast-only. Anyone picking this transport will find out by trying. If you know which of the current adapters would work over it, a short list on that page would save a round of issues. If you would rather not guess, leave it and I will work it out from the registry.

KristianP26:
Rebased onto dev and took both robustness points myself rather than leaving them with you, since they were mine to argue for.

Clock skew now names itself. After twenty advertisements dropped as stale with none delivered, it warns once with the measured offset and points at NTP on both hosts. Two tests: one that the warning fires exactly once on a 600 s offset, one that it stays quiet as soon as anything has been delivered, so a genuine cache replay on subscribe cannot trip it.

A permanent error after the first connection is now logged at error level and says the transport has given up, rather than reading like an attempt that will retry.

I also released the workflow runs, which were sitting on first-time-contributor approval. That is why only GitGuardian had reported.

Green on the rebased branch: tsc, eslint, prettier, 147 test files, 2554 tests.

That leaves only the docs question from my review, and it is not a blocker: the supported-scales table does not mark which adapters are broadcast-only, so someone picking this transport finds out by trying. If you know which of the current adapters would work over it, a short list on the ha-bluetooth page would save a round of issues. If you would rather not, say so and I will work it out from the registry myself.

KristianP26 (approved):
Approved. Rebased, both robustness points taken, full suite green locally. Verified against Home Assistant's own serialize_service_info that the time field is unix epoch, so the stale gate is correct.

gvarph:
Thanks for the rebase, for taking both robustness points, and for releasing the workflow runs. The skew warning firing only while nothing has been delivered is the right shape; a cache replay on subscribe would otherwise have tripped it on every restart.

Docs question: I went through the registry rather than guessing, and pushed one commit, `docs(ha-bluetooth): list which adapters the transport can serve`, that adds a "Which scales work" section to the page and points the existing tip at it.

The rule is simply "declares `parseServiceData` or `parseBroadcast`", which is also what `logTransportCapabilities` prints at startup, so the page and the log cannot drift apart. That gives seven rows:

| Adapter | Over HA Bluetooth |
| --- | --- |
| Xiaomi Mi Scale 2 (MIBCS / MIBFS / XMTZC05HM) | weight + impedance |
| Xiaomi Mi Smart Scale 2 (XMTZC04HM) | weight only |
| Xiaomi Mijia Body Composition Scale S800 | weight only, `bind_key` |
| Xiaomi Body Composition Scale S400 (#374) | weight + impedance + HR, `bind_key` + `scale_mac` |
| Silvergear Smart Scale 108 | weight only |
| Eufy Smart Scale P2 / P2 Pro | weight only, as with the local adapter |
| QN Scale, `AABB` broadcast firmware only | weight only; connectable QN units need GATT |

Everything else on the list has only `charNotifyUuid` and is named in the startup warning. I also added a sentence on why: HA's Bluetooth websocket API only has the four subscribe commands (advertisements, connection allocations, scanner details, scanner state), so an outside client cannot connect or receive notifications even though HA's own integrations can through the same proxies. That should head off "would a better proxy fix it" issues. Two judgement calls you may want to overrule:

- I listed the S400 row even though it lives in #374, since the tip on this page already mentioned it and whichever PR lands second will otherwise leave the page a row short. Drop it if you would rather the page only describe what is on `dev` at the moment it merges.
- QN and Eufy are dual-mode, so the log will call them "broadcast adapters" while a connectable QN unit still cannot be read this way. The table says so in words; I did not touch the log line.

The Xiaomi S400 in front of me has been read over this transport for a week (details in #374), which is the only row in that table I can vouch for from hardware rather than from the code.

KristianP26:
Your follow-up commit made it into the merge, so the "Which scales work" section is on dev. Both judgement calls were right and I am not overruling either.

The S400 row: keep it. Whichever of the two PRs landed second would otherwise have left the page a row short, and it landed anyway, so the page is correct as it stands.

QN and Eufy being dual-mode: the table saying it in words is the right place for that. The startup line calls them broadcast adapters because that is what the capability check can see, and a connectable QN unit still cannot be read this way. Wording that into the log would mean the log making a claim about firmware it has not met yet.

I checked the seven rows against the live registry rather than taking them on trust, filtering for adapters that declare parseServiceData or parseBroadcast, and the list matches exactly: Mi Scale 2, Mi Smart Scale 2, S800, S400, Silvergear 108, Eufy P2, QN. Tying the rule to the same predicate the startup line uses is what makes that checkable at all, which is the part I liked most.

The sentence explaining that the limit is Home Assistant's API and not the proxy hardware will save a round of issues on its own.

Issue #322, opened 2026-08-19
Juniper/Hutbit-compatible scale follow-up: raw impedance decoded, FFB1 command meanings identified, and unit/profile behaviour clarified


RussH:
### Description

I’ve been doing live testing with a **Juniper-branded body-fat scale** which appears to use the same Lefu/Fitdays FFB0 `AC02` protocol as the Hutbit 218008/WL292 / SWAN devices already supported through #254 and #278.

I started from the existing Hutbit support in `ble-scale-sync` and then did additional packet captures and live experiments on the Juniper unit. I also used AI-assisted comparison against openScale and related Fitdays/Lefu implementations to help interpret some of the previously opaque FFB1 commands.

The main result is that the scale appears to provide **valid raw impedance**, and several parts of the command sequence now seem to have identifiable meanings rather than simply being replayed captured packets.

## Observed protocol behaviour

- `AC 02 [weight] 00 00 CE xx` appears to be the live/settling weight stream.
- `AC 02 [weight] 00 00 CA xx` is the stable weight.
- On this Juniper unit, weight remains encoded as big-endian deci-kg regardless of the configured display unit.
- After stable `CA`, if the connection remains open, the scale enters a BIA phase and sends repeated `FD00` frames followed by:

```text
AC 02 FD 01 [impedance_hi] [impedance_lo] CB xx
```

Across several real weigh-ins I have repeatedly received:

```text
AC 02 FD 01 02 06 CB D1
```

which decodes as:

```text
0x0206 = 518 Ω
```

This has been repeatable across measurements with different body weights, so it looks like a genuine raw impedance measurement rather than the unreliable body-fat result previously associated with this protocol family.

`FD FF` also appears to represent an unavailable/failed impedance result, for example when electrode contact is not established.

The existing `completionHoldMs` / `isFinal()` mechanism in `ble-scale-sync` works well for this: stable weight can be treated as complete, but the connection remains open briefly to allow the richer FD01 impedance result to arrive.

## FFB1 command findings

The existing Hutbit implementation replays the captured sequence:

```text
AC 02 FA 01 00 00 CC C7
AC 02 FB 02 1F A5 CC 8D
AC 02 FD E2 01 01 CC AD
AC 02 FC 01 00 00 CC C9
AC 02 FE 06 00 00 CC D0
```

Comparison with related Fitdays/openScale implementations suggests that at least two of these commands have identifiable meanings:

```text
FB <sex> <age> <height_cm>
[...]

PR #383, opened 2026-09-01
feat: add Speediance Smart Scale (FG2211WBF) adapter

chrisauer:
## What

Adds an adapter for the **Speediance Smart Scale (FG2211WBF)** — a Lefu/Icomon
FFB0 sibling of the Robi S9, advertised as `SPEED_S_*`.

On this hardware the existing Robi S9 handshake yields **weight only**: the scale
never runs its impedance phase. The vendor app's handshake includes two frame
types the Robi handshake omits — a `b8` timestamped-identity frame (carrying a
unix timestamp + an `icomon` identity) and a `b4` frame — and those arm the
bioimpedance measurement. With them, the final result arrives as a multi-part
`a7` frame instead of the Robi `a3`.

## How it was derived

From a PacketLogger capture of the iOS app doing a full weigh-in (barefoot +
handles). The 16-frame FFB1 handshake is replayed **verbatim**; the 20-byte
frame checksum is not cracked, but — as with the Robi adapter — the scale
accepts the stale replay. The `a7` part-00 frame carries:

- **weight** — u24 BE grams at offset 9 (verified: `01 2a 3e` = 76.350 kg)
- **whole-body impedance** — u16 LE at offset 15

Per-limb 8-electrode segmental data rides `a7` part-01 and is **not decoded**
here (whole-body BIA does not need it). The impedance offset comes from a single
capture, so the derived body-fat number is plausible but not vendor-exact — the
comments say so, matching the Robi adapter's cautious framing.

Registered at priority 45 so `SPEED_S_*` resolves to this adapter rather than
the weight-only Robi one.

## Tests

- `matches()` / registry resolution (`SPEED_S_*` → Speediance, not Robi)
- handshake replay order + presence of the `b8`/`b4` arming frames
- `a7` weight + impedance parse, and A2-live-frame rejection, from the real capture
- required `registry-collision` fixture

Full scales suite green (960 tests). No new dependencies.


KristianP26:
Thanks for this, and for doing it off a real capture rather than guessing. The framing analysis is right, the priority 45 over Robi's 40 is the correct way to keep a SPEED_S unit off the Robi handshake, and the A7 part-00 weight decode checks out: `01 2a 3e` is 76350 g, which is what your fixture asserts.

Tests pass here and tsc is clean. Three things need sorting before I can merge, and the first one is the important one.

**1. `computeMetrics` does not actually compute BIA**

The comment says "Whole-body BIA from weight + impedance", but `buildPayload` never runs the BIA estimator. Look at its body: `const bodyFatPercent = comp.fat ?? estimateBodyFat(bmi, p)`. With `comp` passed as `{}`, body fat always comes from the Deurenberg BMI estimate and the impedance you worked so hard to unlock is only echoed into the payload.

I checked this rather than eyeballing it. Feeding the helper your fixture weight with impedance 3022, 302.2, 2967 and 296.7 in turn gives identical output every time:

```
   3022 fat% 21.77 water% 57.11 muscle 57.22
  302.2 fat% 21.77 water% 57.11 muscle 57.22
   2967 fat% 21.77 water% 57.11 muscle 57.22
  296.7 fat% 21.77 water% 57.11 muscle 57.22
```

The adapter that does it properly is Hutbit:

```ts
const fat =
  reading.impedance > 0 ? computeBiaFat(reading.weight, reading.impedance, profile) : undefined;
return buildPayload(reading.weight, reading.impedance, { fat }, profile);
```

To be clear about where this came from: you copied Robi S9, and Robi S9 has the same bug with the same misleading comment. That is on me, not on you. I am opening a separate issue for the other adapters affected, so please just fix it here for Speediance.

**2. The impedance value is out of physiological range, so I do not want to publish it yet**

3022 ohm is not a whole-body BIA figure. For comparison, the Hutbit adapter gates impedance to 150-1200 ohm and logs and drops anything outside that, which came out of #322.

What makes me think the field is right but the scaling is not: your A7 frame carries a second u16 LE two bytes later, `97 0b` = 2967, sitting right next to `ce 0b` = 3022. Two values around 3000 next to each other look like a pair of channels, and divided by ten they become 302.2 and 296.7, which are both squarely in range for a whole-body measurement.

That is a hypothesis, not a decode, and this project's rule is that impedance does not ship on a hypothesis. Once point 1 is fixed the number stops being cosmetic and starts driving the body fat percentage the user sees, so a factor of ten here is the difference between a plausible reading and a wrong one.

The cheapest way to settle it: do a weigh-in with the Speediance app open, note the body fat percentage it shows you, and post it here with the DEBUG frame from the same session. Then we can see which divisor reproduces the app.

**3. The adapter counts are not updated**

Adding this adapter takes the registry from 33 to 34, so the protocol adapter count goes 32 to 33. Six places carry that number and all of them still say 32:

- `docs/public/logo.svg`, the text `32.11` (it is adapters dot exporters, so `33.11`)
- `docs/faq.md`, twice: the JSON-LD block on line 12 and the prose on line 67
- `docs/alternatives.md`, line 28
- `docs/guide/supported-scales.md`, line 12
- `CONTRIBUTING.md`, line 52

One more question, not a change request. The handshake embeds what looks like a device identity: `6a 96 d1 e8` in the `b8` frames, and your scale's own A7 replies start `6a 96 d1 ff`. Does the scale accept that replay on a unit that is not yours, or does it only work because the id came from your own device? If you happen to have a second unit to test against, that would be good to know. If not, say so and I will note it as untested rather than hold the PR for it.

KristianP26:
Rebased onto dev and pushed the two points that were mine to fix, so only one thing is still owed from you.

**Fixed on your branch**

The BIA gap is closed. `computeMetrics` now runs `computeBiaFat` when there is an impedance, the way Hutbit does, so the number the scale measures actually reaches the body composition. As I said, that was inherited from Robi S9, not something you did.

The counts are updated across all six places. They moved again while this was open, because the Etekcity ESF-551 adapter landed on dev in the meantime, so it is registry 35 and protocol adapters 34 now. Your docs row is in dev's current table format, which changed under you too.

**The impedance, and why this is now safe to merge without the answer**

I did not guess the divisor. Instead the adapter now gates the impedance to the 150 to 1200 ohm whole-body range the Hutbit adapter established in #322. Your capture's 3022 falls outside it, so it is dropped with a debug line and body composition falls back to the BMI estimate.

That makes the two changes safe together. Without the gate, wiring up the BIA estimator would have started publishing a body fat computed from a number that is plausibly ten times too large, which is worse than the BMI estimate it replaced. With it, the adapter reports weight correctly and simply does not claim an impedance it cannot vouch for.

Your test now asserts `impedance: 0` for that frame, with a comment saying why, and there is a second test with a value inside the range to prove the path works.

**What is still owed**

One weigh-in with the Speediance app open, and the body fat percentage it showed you, alongside the DEBUG frame from the same session. That is the only thing that settles whether the divisor is 1 or 10. If it is 10, the fix is one constant and the impedance starts flowing.

Everything green on the rebased branch: tsc, eslint, prettier, 145 test files, 2517 tests.

Still open from my original review, and genuinely just a question: does that handshake work on a unit that is not yours? It embeds `6a 96 d1 e8`, and your scale's own replies start `6a 96 d1 ff`, so it may be device-specific. If you cannot test it, say so and I will note it as untested rather than hold the PR.

KristianP26 (approved):
Approved. The impedance gate makes this safe to merge without the divisor answer: an out-of-range value is dropped and body composition falls back to BMI, so the adapter never publishes a number it cannot vouch for. The divisor and the handshake portability question are follow-ups, not blockers.

KristianP26 (approved):
Re-approving after the rebase onto current dev and the count bump to 35 protocol adapters, which the Xiaomi S400 merge moved again. Full suite green: 150 test files, 2632 tests.

KristianP26:
Heads up on how this landed, because the PR page will look odd.

Your PR was targeting `main` rather than `dev`, which I did not check before merging. Since I had rebased your branch onto `dev` for the count fix, the squash against `main` swept up the whole difference between the two branches into one commit named after your adapter: 125 files. That is on me, not on you, and it is being unwound on the `main` side.

Your actual work is now on `dev` where it belongs, as three commits with you as author on the first:

```
e1284ad feat: add Speediance Smart Scale (FG2211WBF) adapter
580b18a fix(speediance): run the BIA estimator, gate the impedance, and update the counts
dae289d docs: count 35 protocol adapters after the Xiaomi S400 landed
```

Verified on dev before pushing: tsc, eslint, prettier, 150 test files, 2632 tests.

So the adapter is in and will ship in the next release. The one thing still open is the impedance divisor: one weigh-in with the Speediance app's own body-fat figure alongside the DEBUG frame, and we can decide between 1 and 10. Until then the 150 to 1200 ohm gate drops the 3022 value and body composition falls back to the BMI estimate, which is the honest behaviour rather than a wrong number.

Also still curious whether that handshake works on a unit that is not yours, given it embeds `6a 96 d1 e8`.

PR #402, opened 2026-09-08
fix: secret handling, file permissions, and log injection

KristianP26:
Five findings from a defensive security review of `src/`. None is remotely exploitable on its own; the first is the one that matters.

## 1. `setup --non-interactive` inlined every secret into `config.yaml`

The write-back fired on two conditions: slugs were generated, **or the schema had filled in defaults**. The second case wrote `result.data` - a tree derived from the *resolved* config, with every `${ENV_VAR}` already replaced by its real value.

The Garmin password, the MQTT password, every exporter token the user had deliberately kept in `.env` was inlined into `config.yaml` in plaintext, permanently. The whole point of the `${VAR}` indirection - keeping secrets out of the file people back up and paste into issues - was gone.

It fired on **essentially every hand-written config**, because the schema fills defaults (`device_id`, `topic_prefix`, `out_of_range`, `runtime.*`) that a hand-written file does not carry.

Materialising defaults into the file was only ever cosmetic - the schema applies them on every load regardless - so that branch is dropped rather than repaired. The remaining write-back serialises `parsed`, the tree read from disk, which cannot contain a resolved reference. Side benefit: a run with nothing to change no longer reformats the file and drops the user's comments.

## 2. `atomicWrite` stripped `0600` off `config.yaml` on the first weigh-in

The tmp file was created with no mode, so it landed at `0666 & ~umask` (usually 0644), and the rename carried that mode onto the target. The wizard chmods `config.yaml` to 0600 once, at save time - but `updateLastKnownWeight()` calls `atomicWrite` on **every non-dry weigh-in**. So the first export after setup made the credentials readable by any other local UID, permanently.

The tmp file needs the mode too (same content, predictable name). The EBUSY/EPERM fallback additionally needs an explicit `chmod`, because `writeFileSync` applies `mode` only when it *creates* the file - and that branch exists precisely because the target already exists.

## 3. Heartbeat followed a symlink in `/tmp`

Fixed, world-predictable path in a world-writable directory, rewritten every 30 s with every error swallowed. A local user could pre-create it as a symlink to any file this process can write - `config.yaml`, a crontab, an `authorized_keys` - and each tick would follow it and truncate the target.

The path cannot move (the Docker HEALTHCHECK reads it by name), so the open now carries `O_NOFOLLOW`; an ELOOP simply skips the tick.

## 4. Log injection from an attacker-controlled BLE name

The local name arrives unfiltered from anyone with a radio in range, or anyone who can publish to a proxy topic, and **no producer in the tree filtered it**. It is logged for every advertisement seen during a scan, matched or not, so nothing gates it.

A crafted name carrying CR/LF forges convincing log lines in journald or docker logs - exactly what reporters are asked to paste into public issues - and ANSI/OSC escapes rewrite the terminal of whoever `cat`s the file.

`safeName()` hex-escapes control characters rather than stripping them, so a name that really does contain one stays distinguishable. Applied at all **eleven** sites across the four transports. Adapter names are our own constants and are left alone.

## 5. Strava OAuth logged the raw upstream body

Hardening, not a live leak: Strava does not echo `client_secret` back today. But this is the one path that has just transmitted it, and an unbounded verbatim dump of a third party's response body is not something to leave to their discretion. Same cap and reasoning as `notification-message.ts`.

## Scope of #2 and #3

Both are reachable only by another local user on a shared host. A single-user Pi, or a container with a private `/tmp`, is unaffected.

## Verification

The secret-leak test was run against the unfixed code and **fails** there. The file-mode assertions skip on Windows and run in CI on Linux.

- `npx tsc --noEmit`, `npm run lint`, `npm run format:check` clean
- `npm test`: 159 files, 2814 passing / 3 skipped

### DIFF
--- a/src/ble/advertisement.ts
+++ b/src/ble/advertisement.ts
@@ -236,6 +236,27 @@ function blob(data: Buffer): string {
 }
 
 /**
+ * Render a device-supplied name safe to put in a log line.
+ *
+ * The local name comes straight off the air, unfiltered, from anyone with a
+ * radio in range - or from anyone who can publish to a proxy topic. It is
+ * logged for EVERY advertisement seen during a scan, matched or not, so no
+ * pairing or trust step gates it. A crafted name carrying CR/LF forges
+ * convincing log lines in journald or docker logs, which is exactly what
+ * reporters are asked to paste into public issues, and ANSI/OSC escapes rewrite
+ * the terminal of whoever cats the file.
+ *
+ * Control characters are hex-escaped rather than stripped so a name that really
+ * does contain one stays distinguishable from a name that does not.
+ */
+export function safeName(name: string | undefined): string {
+  if (!name) return '';
+  return name.replace(/[\x00-\x1f\x7f]/g, (c) => {
+    return `\\x${c.charCodeAt(0).toString(16).padStart(2, '0')}`;
+  });
+}
+
+/**
  * One-line summary of everything an adapter's `matches()` is allowed to see.
  *
  * Adapter mis-routing was the root cause of #317, #318 and #319, and every one
@@ -252,7 +273,7 @@ export function formatAdvert(address: string, info: BleDeviceInfo): string {
   // to mean anything.
   const parts = [
     `[${address.toUpperCase()}]`,
-    `name=${info.localName ? `"${info.localName}"` : '(none)'}`,
+    `name=${info.localName ? `"${safeName(info.localName)}"` : '(none)'}`,
   ];
   parts.push(`uuids=${uuidList(info.serviceUuids)}`);
   if (info.manufacturerData) {
--- a/src/ble/advertisement.ts
+++ b/src/ble/advertisement.ts
@@ -246,13 +246,25 @@ function blob(data: Buffer): string {
  * reporters are asked to paste into public issues, and ANSI/OSC escapes rewrite
  * the terminal of whoever cats the file.
  *
- * Control characters are hex-escaped rather than stripped so a name that really
- * does contain one stays distinguishable from a name that does not.
+ * Escaped rather than stripped, so a name that really does contain a control
+ * character stays distinguishable from one that does not. That only holds if
+ * the escape character is escaped too - otherwise a name containing the four
+ * literal characters `\x0a` renders identically to one containing a real LF -
+ * so backslash is in the class.
+ *
+ * The class covers more than C0. U+0080..U+009F are the C1 controls, and
+ * xterm-family terminals honour U+009B and U+009D as CSI and OSC introducers
+ * even in UTF-8 mode; a BLE name is UTF-8 decoded by every stack, so those
+ * arrive intact. U+2028 and U+2029 are line breaks to anything that splits on
+ * Unicode newlines rather than on LF.
  */
 export function safeName(name: string | undefined): string {
   if (!name) return '';
-  return name.replace(/[\x00-\x1f\x7f]/g, (c) => {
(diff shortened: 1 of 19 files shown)


KristianP26:
Independent review found the `safeName` rollout incomplete. **"All eleven sites" was wrong** - eight more were unwrapped, and the worst miss was `npm run scan`, which prints every device name and is the output users are told to paste into issues and to read their `scale_mac` from. `diagnose` and the wizard device picker had it too, and the wizard writes to a live terminal, so an escape sequence there needs no log file at all.

Fixed in 1e820d7:

- Escape once where each transport builds `ScanResult.name`, so `scan`, `diagnose` and the wizard inherit it instead of needing four more call-site wraps. The two `Device found (...) but no adapter recognized it` Errors are wrapped directly, since `run.ts` logs `err.message` verbatim.
- **Widened the class past C0.** U+0080..U+009F are the C1 controls, and xterm-family terminals honour U+009B (CSI) and U+009D (OSC) even in UTF-8 mode - a BLE name is UTF-8 decoded by every stack, so those arrived intact through a C0-only class. Added U+2028/U+2029 and backslash; without escaping the escape character the "stays distinguishable" property the doc comment claimed did not hold.
- Corrected two comments that overstated their case: the heartbeat symlink note (Debian and Raspberry Pi OS default to `fs.protected_symlinks`, Docker and the add-on get a private `/tmp`, so this is defence in depth, not a live truncation primitive) and the Strava cap, which claimed to match `notification-message.ts` at 200 when that file caps at 120.
- The EBUSY fallback `chmod` now has a test, with an explicit `expect(spy).toHaveBeenCalled()` - without it the mock failing to bind would let the rename succeed, the mode would come from the 0600 tmp file, and the assertion would pass while the branch never ran.

Not changed, deliberately: the unvalidated `JSON.parse` on the mqtt-proxy scan topic and the comment-dropping YAML round-trip on the slug path are both pre-existing and filed separately (#404, #407).

`tsc`, lint, prettier clean; 159 files / 2817 passed / 4 skipped.

KristianP26:
Blocking finding fixed, plus `dev` merged in.

**`src/ble/handler-node-ble/scan.ts` built `ScanResult.name` unescaped** - and node-ble is the default transport on Linux/RPi, so that is where most users are. My "rollout complete" claim was wrong for the second time. Escaped at the construction site like the other four transports.

It survived two review rounds because **no test exercises node-ble `scanDevices` at all**. Rather than write one unit test for the transport that happened to be missed, `tests/ble/scan-result-name-escaped.test.ts` now pairs `name:` with its sibling `matchedAdapter:` across `src/ble/`, finds all five ScanResult producers, and requires every one to escape. It asserts the producer count too, so it cannot go vacuous, and it fails for any transport added later that forgets. Verified: reverting the node-ble line fails exactly that file's case.

Also resolved the predicted conflict with #400 in `handler-esphome-proxy/scan.ts` - adjacent import lines, kept both additions (`safeName` from `../advertisement.js`, `withIdleTimeout` from `../types.js`).

Not changed, noted as follow-ups: the raw `ESP32 status: ${payload}` log line and the unescaped `normalizeUuid`/`formatMac` paths reach a log only from a broker-level attacker who can already forge readings; bidi and zero-width characters are spoofing rather than terminal control; `O_NOFOLLOW` does not cover a planted hard link, which is what `fs.protected_hardlinks` is for.

Gate after the merge: `tsc`, lint, prettier clean; 2850 passed / 5 skipped (the skips are the Windows-only file-mode and O_NOFOLLOW assertions, which run on the Linux CI).

Issue #206, opened 2026-05-21
feat(exporter): add Fitbit exporter


KristianP26:
## Summary

Add an exporter that logs body-composition readings to [Fitbit](https://www.fitbit.com) via the Fitbit Web API.

Fitbit reaches a large mainstream userbase outside the Garmin/Strava endurance crowd, broadening the project's audience.

## API

- **Auth:** OAuth2, scope `weight`.
- **Weight:** `POST /1/user/-/body/log/weight.json` with `weight`, `date`, `time`.
- **Body fat:** `POST /1/user/-/body/log/fat.json` with `fat`, `date`, `time`.
- **Rate limit:** 150 requests/hour per user — fine for scale readings, but handle 429s gracefully.
- Docs: https://dev.fitbit.com/build/reference/web-api/body/

## Implementation notes

- New `src/exporters/fitbit.ts` implementing the `Exporter` interface.
- Config: OAuth client credentials + token (with refresh). Add `FitbitConfig` to `config.ts` and wire through `ExporterName`, `KNOWN_EXPORTERS`, `createExporters()`, registry/schemas.
- `supportsBackdate: true` — endpoints accept explicit `date`/`time`, so honour `context.timestamp`.
- OAuth flow in the wizard (compare with `strava-setup.ts`).
- Fitbit only exposes weight + body fat — other `BodyComposition` metrics are dropped; document this.
- Implement `healthcheck()` for the wizard connectivity test.
- Docs page under `docs/exporters`.

## Acceptance criteria

- [ ] Live reading appears in Fitbit (weight + body fat)
- [ ] Historical/back-dated readings land on their original date/time
- [ ] OAuth setup works in the wizard with token refresh
- [ ] 429 rate-limit responses handled without crashing
- [ ] `healthcheck()` validates credentials
- [ ] Docs + config example updated

KristianP26:
Closing this as blocked by the platform rather than by effort.

The two endpoints this issue is built on are going away. Fitbit's own reference page for the Body endpoints now carries the banner:

> The Fitbit Web APIs are moving to a new, scalable infrastructure. We will be deprecating the legacy Fitbit Web API in September 2026.

That covers the whole legacy Web API, including `POST /1/user/-/body/log/weight.json` and `POST /1/user/-/body/log/fat.json`. Building against them now would ship an exporter that breaks within weeks.

[...]

PR #414, opened 2026-09-09
feat: add Google Health exporter

RussH:
## Summary

Adds a standalone Google Health exporter, allowing BLE Scale Sync to upload weight measurements and, optionally, derived body-fat percentage to the Google Health API.

The exporter is intentionally independent of the scale/protocol work and is configured per user, as the OAuth refresh token authorizes access to an individual Google Health account.

## Changes

- Add new `google-health` per-user exporter
- Upload weight measurements to the Google Health API as actively measured scale data
- Optionally upload valid derived body-fat percentage
- Add OAuth 2.0 refresh-token authentication with:
  - in-memory access-token caching
  - expiry handling
  - coalesced concurrent token refreshes
  - one forced token refresh/retry following an HTTP 401
- Avoid generic retries for health-data POST requests to reduce the risk of duplicate data points
- Support historical/backdated readings using the reading timestamp
- Add optional scale manufacturer and display-name metadata
- Add Google Health configuration to the exporter registry and legacy environment-variable configuration
- Handle `write_body_fat` through the existing boolean parsing safeguards, including `${ENV_VAR}` values such as `"false"`
- Add exporter and registry tests, including OAuth, token caching, error handling, body-fat behaviour, healthcheck and configuration handling
- Add Google Health setup and `config.yaml` documentation, including the required OAuth scope and refresh-token setup
- Document that Google Health must be configured under a user's `exporters:` list rather than `global_exporters`
- Update published exporter counts from 11 to 12

## Test Plan

- [ ] Tests pass (`npm test`) — the Google Health changes pass, but 8 existing `tests/ble/handler-node-ble/scan-order.test.ts` tests time out on this machine. The identical 8 failures were reproduced against a clean detached current `origin/dev`, independently of this branch.
- [x] TypeScript compiles (`npm run build`)
- [x] Tested manually

Manual testing included completing the OAuth flow with Google Health and successfully uploading real weight and body-fat measurements. Refresh-token exchange and subsequent access-token use were also verified end-to-end.

Dedicated Google Health and exporter-registry tests pass, as do the documentation build and published exporter-count checks.

## Notes

This PR contains no BLE scale/protocol changes.

Google Health OAuth credentials are per-user and should normally be supplied through `${ENV_VAR}` references rather than committed directly to `config.yaml`.

The healthcheck validates OAuth access-token refresh without writing a synthetic health measurement.

The 8 local `scan-order.test.ts` timeouts are unrelated to this PR and reproduce unchanged on current `origin/dev`.

### DIFF
--- a/README.md
+++ b/README.md
@@ -91,7 +91,7 @@ Requires Node.js v22+ and a BLE adapter. See the **[full install guide](https://
 ## Features
 
 - **[25+ scale brands](https://blescalesync.dev/guide/supported-scales).** Xiaomi (Mi Scale 2 passive broadcast), Renpho (Elis 1, ES-CS20M, QN-Scale), Eufy, Yunmai, Beurer (incl. BF720), Sanitas (incl. SBF70 body composition), Medisana, and more.
-- **[11 export targets](https://blescalesync.dev/exporters).** Garmin Connect, Strava, Intervals.icu, Runalyze, Wger, MQTT (Home Assistant), InfluxDB, Webhook, Ntfy, Telegram, File (CSV/JSONL).
+- **[12 export targets](https://blescalesync.dev/exporters).** Garmin Connect, Strava, Intervals.icu, Runalyze, Wger, MQTT (Home Assistant), InfluxDB, Webhook, Ntfy, Telegram, File (CSV/JSONL).
 - **[10 body metrics](https://blescalesync.dev/body-composition).** BIA-based body composition from weight + impedance.
 - **[Multi-user](https://blescalesync.dev/multi-user).** Automatic weight-based identification with per-user exporters.
 - **Historical sync.** Replays a scale's onboard cache of offline measurements with their original timestamps to exporters that support back-dating (Garmin Connect, InfluxDB, File).
--- a/docs/exporters.md
+++ b/docs/exporters.md
@@ -1,6 +1,6 @@
 ---
 title: Exporters
-description: Configure Garmin Connect, Strava, Intervals.icu, Runalyze, Wger, MQTT, Webhook, InfluxDB, Ntfy, Telegram, and File export targets.
+description: Configure Garmin Connect, Google Health, Strava, Intervals.icu, Runalyze, Wger, MQTT, Webhook, InfluxDB, Ntfy, Telegram, and File export targets.
 head:
   - - meta
     - name: keywords
@@ -9,7 +9,7 @@ head:
 
 # Exporters
 
-BLE Scale Sync exports body composition data to 11 targets. The [setup wizard](/guide/configuration#setup-wizard-recommended) walks you through exporter selection, configuration, and connectivity testing.
+BLE Scale Sync exports body composition data to 12 targets. The [setup wizard](/guide/configuration#setup-wizard-recommended) walks you through exporter selection, configuration, and connectivity testing.
 
 Exporters are configured in `global_exporters` (shared by all users). For multi-user setups with separate accounts, see [Per-User Exporters](/multi-user#per-user-exporters). All enabled exporters run in parallel; the process reports an error only if **every** exporter fails.
 
@@ -22,6 +22,7 @@ Exporters are configured in `global_exporters` (shared by all users). For multi-
 | [**Ntfy**](#ntfy)               | Push notifications to phone/desktop                    |
 | [**Telegram**](#telegram)       | Send measurement notifications to a Telegram chat      |
 | [**File (CSV/JSONL)**](#file)   | Append readings to a local file                        |
+| [**Google Health**](#google-health) | Push weight and derived body fat to Google Health  |
 | [**Strava**](#strava)           | Update weight in your Strava athlete profile           |
 | [**Intervals.icu**](#intervals) | Push weight + body fat to Intervals.icu wellness       |
 | [**Runalyze**](#runalyze)       | Push weight + body composition to Runalyze metrics     |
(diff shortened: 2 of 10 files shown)


KristianP26:
Thanks for this, and for scoping it cleanly away from the BLE side.

I checked the API surface against the live v4 discovery document (revision 20260908) rather than the reference pages, since that is what the service accepts, and it holds up: `users/me` and the hyphenated `body-fat` both appear there verbatim, `weightGrams`, `percentage` and `sampleTime` match the schemas, `physicalTime` and `utcOffset` are both `Required` in `ObservationSampleTime` while the output-only `civilTime` is correctly left out, and `SCALE`, `ACTIVELY_MEASURED` and `DERIVED` are all valid enum members. `supportsGlobal: false` with `supportsPerUser: true` is the right call for an account tied to one person, and `supportsBackdate: true` is earned rather than assumed, since `context.timestamp` really is what ends up in `sampleTime`.

The suite is green here on your branch: 176 files, 3030 tests. The 8 `scan-order.test.ts` timeouts you hit do not reproduce on this machine on your branch or on `dev`, so that looks like your environment rather than anything you changed.

Two things block the merge.

## 1. The retry queue will duplicate weight points, which is the failure you designed against

You deliberately keep the create out of `withRetry` because it is not idempotent, and that is correct: the method takes `parent` and nothing else, so there is no client-supplied id for the server to deduplicate on.

The problem is one level up. `export()` returns `{success: false}` after the weight point has already been written, whenever the body-fat POST is the thing that fails. With `supportsBackdate: true` and `runtime.retry_failed_exports` (default `true`, in `dev` since v1.29.0), `queueFailedExports` in `src/runtime/processor.ts:99` persists the reading and `flushQueue` in `src/runtime/export-queue.ts:190` replays the whole `export()`, up to `MAX_ATTEMPTS = 5`. So the exact case your test "reports partial failure when weight succeeds but body fat fails" pins down ends with up to six identical weight points in Google Health. A connection that drops after Google committed the write does the same thing.

`src/runtime/export-queue.ts:132` already names `file` and `runalyze` as the targets that must not be redelivered. Either `google-health` joins them, or `export()` stops reporting failure once the weight write has landed, for example by tracking which of the two points succeeded so a replay only sends the missing one.

## 2. The legacy `.env` path is wired into a branch nothing calls

`src/exporters/index.ts` has a `createExporters()` case for `google-health`, but the only caller left is `tests/exporters/index.test.ts`. The live env path is `loadEnvConfig()` -> `createExporterFromEntry()`, and `src/config/env-load.ts` rebuilds each entry with a per-name `Object.assign` chain that has no `google-health` branch.

Result, with all three documented variables set:

```
EXPORTERS=google-health
GOOGLE_HEALTH_CLIENT_ID / _CLIENT_SECRET / _REFRESH_TOKEN set

global_exporters: [{"type":"google-health"}]
Error: Exporter "google-health" is missing required field "client_id". Check your config.yaml.
```

`loadExporterConfig()` reads and validates all three, then the values are dropped before anything can use them, and the user is told to check a `config.yaml` they are not using. The fix is the mapping block in `env-load.ts` alongside `wger` and the rest.

## 3. Body fat is never marked `DERIVED`

`BodyFatDataPoint` carries no `dataSource` at all, so `'DERIVED'` in the `RecordingMethod` union is dead. The field description in the schema, the `writeBodyFat` doc comment and the docs page all say the point is uploaded as derived. Google makes `dataSource` optional, so this fails quietly.

It matters more here than the wording suggests. When a scale reports no impedance, `bodyFatPercent` is a Deurenberg estimate from BMI, not a measurement, and writing it with no recording method makes it indistinguishable from an actively measured value in the user's own health record. Marking these `DERIVED` is the honest representation. The configured `device_manufacturer` and `device_display_name` are dropped from body-fat points as well, while weight points get them.

## Smaller things

- `src/exporters/registry.ts:284-285` casts `config.device_manufacturer` and `device_display_name` with `as string | undefined`. The entry schema is `.passthrough()`, so `device_manufacturer: 12345` in `config.yaml` reaches `buildDataSource`, which calls `.trim()` on it and throws at export time rather than at startup. Everything else in that registry goes through `requireField`, `optionalBool` or `optionalNumber`.
- `README.md` and `docs/index.md` now say 12 export targets, but both lists underneath still enumerate 11 names and omit Google Health. `npm run check:counts` only verifies the number, so this one cannot be caught mechanically.
- `config.yaml.example` has a commented block for every other exporter. Google Health should get one too, under a user's `exporters:`, with the credentials as `${ENV_VAR}` references.
- `getAccessToken(true)` clears the cached token and then returns the in-flight `tokenRefreshInFlight` promise, which can resolve to the token that produced the 401. Unreachable today, since the reading loop and the queue flush are serialized and each user gets its own instance, but it defeats the point of the forced refresh if a concurrent caller is ever added.
- The branch is based on `c51e350`, which predates v1.29.0. It will need a rebase onto `dev`.

## One question before this can ship

I closed #206 (Fitbit) on 2026-08-26 with the reopen trigger "Google grants third-party clients write access to weight and body-fat data types". The write methods were already visible in the discovery document at that point, so their presence is not by itself the evidence.

You say you completed the OAuth flow and uploaded real measurements, which would be that evidence, so I want to record how. What was the verification state of the OAuth client you used? Health data is a restricted scope, so I want to know whether it worked because your Google account was a test user on an unverified client, or because the client actually passed verification, and if the latter, how long it took and whether a security assessment was required.

That answer decides what the setup guide says. If every user has to create their own Cloud project and OAuth client and add themselves as a test user, that belongs at the top of the page as the main step, not as a footnote after the refresh token instructions.

